PR #1, opened 2024-11-05
[WEB-3232] - Configuration for Prisma ORM schema and DB connection

RavenHursT:
https://goodparty.atlassian.net/browse/WEB-3232


.env.example
```diff
@@ -0,0 +1,2 @@
+DATABASE_URL="postgresql://postgres:postgres@localhost:5432/gpdb?schema=public"
```
talmog, line 1: I think I have one more env variable already there

prisma/schema.prisma
```diff
@@ -12,3 +12,37 @@ datasource db {
...
+  candidateContentPrompts
+  blogArticles
+  termsOfService
+}
+
+model Content {
+  createdAt BigInt      @map("created_at")
+  updatedAt BigInt      @map("updated_at")
+  id        String      @id
+  type      ContentType
```
talmog, line 46: where do we add an index to "type"

talmog (changes requested):

.env.example
```diff
@@ -0,0 +1,2 @@
...
+LOG_LEVEL="debug" # Available levels: https://github.com/pinojs/pino/blob/main/docs/api.md#level-string
```
tomer-tgp, line 2: I think I have an env variable already there

prisma/schema.prisma
tomer-tgp, line 46: can we add an index here?

tomer-tgp (changes requested):

.env.example
RavenHursT: huh?

What do you mean? 🤔

prisma/schema.prisma
RavenHursT: On the type?

Does that make sense? The types here aren't unique.

When I mentioned we could do that, I was talking about adding an index on unique lookup properties on the JSONb `data` column (i.e. 'slug' for `blogArticle` content records).  But also, I also was saying we _can_ do that, if we find we need it.

ATM we don't have that many records in Contentful that I think adding an index here would be perceptible, do you think?

.env.example
tomer-tgp: oh I am sorry it is git ignored. I added this to .env
```
CORS_ORIGIN=http://localhost:4000
```

prisma/schema.prisma
tomer-tgp: it is the main way we access this record, and contentful keeps growing. It will reduce the lookup time from O(n) to O(logN), so why not?
the column doesn't have to be unique to be indexed; this is not a primary key, of course.

tomer-tgp (approved):

PR #4, opened 2024-11-08
Web 3340 jobs

tomer-tgp:
Migrate jobs.
Notice I removed the `api` before the `v1`
http://localhost:3000/v1/jobs
http://localhost:3000/v1/jobs/1c58c415-9699-4b7a-996b-811d235128dc


src/jobs/jobs.controller.spec.ts
```diff
@@ -0,0 +1,20 @@
+import { Test, TestingModule } from '@nestjs/testing'
```
tomer-tgp, line 1: Should we keep these?  we can remove them since we plan to test via postman

src/jobs/jobs.service.ts
```diff
@@ -0,0 +1,48 @@
+import { Injectable } from '@nestjs/common'
+import axios from 'axios'
+import { config } from 'dotenv'
+
+config()
```
tomer-tgp, line 5: this is probably why we need a configurationModule. We need to do this every time we try to access an env variable

RavenHursT: I saw this.. 

I could've sworn as long as `config()` is run once in the `main.ts` file, the `process.env` values _should_ be available everywhere.. 🤔

src/jobs/jobs.service.ts
```diff
@@ -0,0 +1,48 @@
...
+    headers: {
+      'Content-Type': 'application/json',
+      Accept: 'application/json',
+      Authorization: `Basic ${Buffer.from(ASHBEY_KEY + ':').toString(
+        'base64',
+      )}`,
+    },
+  })
+  return response?.data?.results
+}
```
RavenHursT, line 48: Looks like the more idiomatic way of doing server-to-server requests in Nest is to use the HttpModule, w/ it's injected service and to have the ... Not to use `axios` directly...

https://docs.nestjs.com/techniques/http-module#getting-started

RavenHursT (changes requested):

src/jobs/jobs.service.ts
RavenHursT: oooooh.. NestJS is loading the service files _before_ it's loading `main.ts`!

```
jobs.service.ts
process.env.LOG_LEVEL => debug
main.ts
process.env.LOG_LEVEL => debug
```

There's gotta be a way to do this somewhere that will run first, before everything else.

src/jobs/jobs.controller.spec.ts
RavenHursT: I took care of these in my last PR... pull from `develop`

https://github.com/thegoodparty/gp-api/pull/5

src/jobs/jobs.service.ts
```diff
@@ -0,0 +1,55 @@
...
+  constructor(private readonly httpService: HttpService) {}
+
+  async findAll() {
+    try {
+      const jobs = await this.fetchJobs('list', { listedOnly: true })
+      return jobs || []
+    } catch (error) {
+      console.error('Error during fetch:', error.message)
+      return []
+    }
```
RavenHursT, line 24: I really think the API should bubble errors like this upward to result in something meaningful, not just an empty response.

For example.. if something's wrong w/ this server-to-server API call, and we just return `200` w/ an empty `[]` in the body, then the client will just render an empty page/list and we won't _know_ something's wrong in PROD until someone bothers to visit the jobs-listing page and then we have a production bug on our hands w/ a 🔥 we gotta put out.

It's typically better to have the mechanisms in place to be proactive, as opposed to reactive, to these types of errors.

My preference would be to log the error using the idiomatic NestJS pattern for the logger: https://konfigthis.com/sdk/#:~:text=Pricing,SDKs%20you%20use.

Then allow this error to be `throw`n up to the caller method in the controller, and then, based on the error, return a `502 Bad Gateway`, since our API is acting as a gateway for the Ashby service behind it.  (Similar to what you did here: https://github.com/thegoodparty/gp-api/pull/4/files#diff-e657b2661a51df80ffee31a7065014c030edfdf7534cb0ab6a0f29d2234aee99R17 albeit, not exactly the same)

This way, the _front-end_ app can then properly introspect the response code, and act accordingly (i.e. throw a client error that FullStory will pick up and let us know something's not right)

RavenHursT (changes requested):

src/jobs/jobs.controller.ts
```diff
@@ -13,21 +13,27 @@ export class JobsController {
...
-    return jobs
   }
 
   @Get(':id')
   async findOne(@Param('id') id: string) {
-    const job = await this.jobsService.findOne(id)
-    if (!job) {
-      throw new NotFoundException(`Job with id ${id} not found`)
+    try {
+      const job = await this.jobsService.findOne(id)
```
RavenHursT, line 34: Didn't you say that Axios throws on a 404 response? 

Won't this just throw, and then get caught and then respond with the 502 bag gateway below? 🤔

RavenHursT (changes requested):

src/jobs/jobs.controller.ts
RavenHursT: Looks like it does 😕

https://github.com/axios/axios/blob/d032edda08948f3f613ff34f32af49cad3aa74db/lib/defaults/index.js#L148

PR #7, opened 2024-11-13
Ravenhurst/web 3321 content module implementation

RavenHursT:
https://goodparty.atlassian.net/browse/WEB-3321

Tests run: https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/34024107-8a81980f-3eb7-4793-86b6-b4a695f5703b


package.json
```diff
@@ -26,7 +26,7 @@
     "@nestjs/platform-fastify": "^10.4.6",
     "@nestjs/swagger": "^8.0.2",
     "@prisma/client": "^5.22.0",
-    "axios": "^1.7.7",
```
tomer-tgp, line 29: why did you remove axios?
https://docs.nestjs.com/techniques/http-module
the docs instruct to install both nextjs/axios and axios

src/contentful/contentful.service.ts
```diff
@@ -0,0 +1,34 @@
...
+  ): Promise<[entries: Entry[], deletedEntries: DeletedEntry[]]> {
+    const {
+      entries,
+      deletedEntries,
+      nextSyncToken: newToken,
+    } = await contentfulClient.sync({
+      ...(initial || !nextSyncToken ? { initial: true } : { nextSyncToken }),
+    })
+    nextSyncToken = newToken
+    return [entries, deletedEntries]
```
tomer-tgp, line 27: I would prefer to return here an object
```
{ entries, deletedEntries }
```
so it is clear what is return and we don't need to know that the first object is the entries and the second one is the deleted entries.

src/content/content.service.ts
```diff
@@ -1,28 +1,119 @@
...
+  async syncContent(seed: boolean = false) {
+    const [entries = [], deletedEntries = []] =
+      await this.contentfulService.getSync(seed)
+    const recognizedEntries = entries.filter(
+      (entry) => CONTENT_TYPE_MAP[entry.sys.contentType.sys.id],
+    )
+    const entryIds = new Set(recognizedEntries.map((entry) => entry.sys.id))
+    const existingContentIds = await this.getExistingContentIds()
+    const existingEntries = intersection(existingContentIds, entryIds)
+    const newEntryIds = difference(entryIds, existingContentIds)
```
tomer-tgp, line 68: why not use the build in function for sets?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/difference

src/content/content.service.ts
```diff
@@ -1,28 +1,119 @@
...
-    return `This action removes a #${id} content`
+  async syncContent(seed: boolean = false) {
+    const [entries = [], deletedEntries = []] =
+      await this.contentfulService.getSync(seed)
+    const recognizedEntries = entries.filter(
+      (entry) => CONTENT_TYPE_MAP[entry.sys.contentType.sys.id],
+    )
+    const entryIds = new Set(recognizedEntries.map((entry) => entry.sys.id))
+    const existingContentIds = await this.getExistingContentIds()
+    const existingEntries = intersection(existingContentIds, entryIds)
```
tomer-tgp, line 67: why not use the build in function for sets?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/intersection

tomer-tgp (changes requested):

src/content/content.service.ts
RavenHursT: Duh.. forgot that was a thing.. 🤦‍♂️

package.json
RavenHursT: I noticed we weren't improving it anywhere and so I removed it and your endpoints still operate as expected? 🤷‍♂️

Pull my branch and see if you get some kind of error maybe? Worked for me.

RavenHursT: Maybe they want it installed because they only have it as a devDependency? Maybe it won't work with a production build? 🤔

https://github.com/nestjs/axios/blob/master/package.json#L32

src/content/content.service.ts
RavenHursT: Turns out that the Set methods aren't available until Node 22... it was released in April, but it's not LTS yet.  We're running on 20.x right now which is the latest LTS..
Should we go ahead and make the new API Node 22?

image
image

PR #9, opened 2024-11-20
[WEB-3322] - get CMS content

RavenHursT:
This is just an initial PR to show the patter while we continue to implement the rest of the transformers for rest of the different Contentful content types that we ingest. 

Successful test run: https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/34024107-aa12c3a8-2c43-41a7-a21e-86f8472ed1ea


api/src/content/content.types.ts
```diff
@@ -0,0 +1,185 @@
...
+import { Content } from '@prisma/client'
+
+// Pulled from https://github.com/lbenie/reading-time-estimator/blob/a5c57383575d8dda37f4d501c57f9e7658db6f0a/lib/reading-time-estimator.ts#L7-L20
+//  Since the author didn't seem to think exporting the type was necessary, I had to copy it here 😡
+type ReadingTime = {
+  readonly minutes: number
+  readonly words: number
+  readonly text: string
+}
+
```
tomer-tgp, line 12: the new blog design is not using "reading time" you can remove this

api/src/shared/util/maps.util.ts
```diff
@@ -0,0 +1,8 @@
+export const mapToObject = (map: Map<string, any>): { [key: string]: any } =>
```
tomer-tgp, line 1: why not replace this function with this?
```
const mapToObject = (map: Map<string, any>): { [key: string]: any } => 
  Object.fromEntries(map);
```

Not only it is more readable, it is also more efficient (since your original function uses the spread operator (...obj) inside a reduce callback, creating a new object on every iteration)

tomer-tgp (changes requested):

api/src/content/content.types.ts
RavenHursT: stone-cold-oh-hell-yeah

api/src/shared/util/maps.util.ts
RavenHursT: Oh wow.. didn't realize we could create objects from Maps in that way.. #TIL Way better! 💯 

image

tomer-tgp (approved):

PR #10, opened 2024-11-21
WEB-3356 WIP

oldfashionedrobot:


api/src/main.ts
```diff
@@ -28,6 +28,13 @@ const bootstrap = async () => {
         : {}),
     }),
   )
+  // app.useGlobalPipes(
```
oldfashionedrobot, line 31: I commented this out for now, if we want to stick with the DTO class validator pattern, we could put this `ValidationPipe` at the global level to not have to add it at every controller

api/src/contentful/contentful.service.ts
```diff
@@ -1,7 +1,10 @@
 import { createClient, DeletedEntry, Entry } from 'contentful'
 import { Injectable } from '@nestjs/common'
 
-const { CONTENTFUL_SPACE_ID, CONTENTFUL_ACCESS_TOKEN } = process.env
+const { CONTENTFUL_SPACE_ID, CONTENTFUL_ACCESS_TOKEN } = process.env as Record<
```
oldfashionedrobot, line 4: Had to add some types/casting here after adding the "strict": true,
"strictNullChecks": true, to the tsconfig

api/src/jobs/jobs.controller.ts
```diff
@@ -19,13 +19,15 @@ export class JobsController {
...
       return await this.jobsService.findAll()
     } catch (e) {
-      this.logger.log(
-        `Error at jobController findAll. e.message: ${e.message}`,
-        e,
-      )
-      throw new BadGatewayException(
-        e.message || 'Error occurred while fetching jobs',
-      )
+      if (e instanceof Error) {
```
oldfashionedrobot, line 22: Had to add typechecks here after adding "strict": true, "strictNullChecks": true, to tsconfig

api/src/shared/helpers/objectHelper.ts
```diff
@@ -0,0 +1,29 @@
...
+        result[key] = deepMerge(obj1[key], obj2[key])
+      } else {
+        // Otherwise, overwrite the value from obj1 with the value from obj2
+        result[key] = obj2[key]
+      }
+    }
+  }
+
+  return result
+}
```
RavenHursT, line 29: Can we please move this to `/util/objects.util.ts` like I've done here for some `Map` util functionality? 🙏 

https://github.com/thegoodparty/gp-api/pull/9/files#diff-0e01814536c3a43f99f91a08ba84cc465f051550f1a248a89a60838c762be414R1-R8

api/src/campaigns/campaigns.service.ts
```diff
@@ -0,0 +1,372 @@
...
+    primaryElectionDateStart,
+    primaryElectionDateEnd,
+    campaignStatus,
+    generalElectionDateStart,
+    generalElectionDateEnd,
+    p2vStatus,
+  })}
+  ORDER BY c.id DESC;
+`
+}
```
RavenHursT, line 347: Doesn't _have_ to be done now.. But, can we add some `TODO`s here to look into seeing if it 's possible to use Prisma's query builder for these now? Waterline didn't play nicely w/ `JSONb` columns, which was why we dropped down toe SQL for these.

But it looks like Prisma can handle JSONb fields just fine in it's query builder: https://www.prisma.io/docs/orm/prisma-client/special-fields-and-types/working-with-json-fields

api/src/campaigns/schemas/campaignList.schema.ts
```diff
@@ -0,0 +1,20 @@
...
+  id: z.coerce.number().optional(),
+  state: isState()
+    .transform((val) => val.toUpperCase())
+    .optional(),
+  email: z.string().email().optional(),
+  slug: z.string().optional(),
+  level: z.enum(['LOCAL', 'CITY', 'COUNTY', 'STATE', 'FEDERAL']).optional(),
+  primaryElectionDateStart: z.string().date().optional(),
+  primaryElectionDateEnd: z.string().date().optional(),
+  campaignStatus: z.enum(['active', 'inactive']).optional(),
```
RavenHursT, line 14: Good to put any string literal into a constant, just in case it gets used elsewhere and then can be managed in just one place.

api/src/shared/validations/isState.ts
```diff
@@ -0,0 +1,23 @@
...
+ * @example
+ * const schema = z.object({
+ *   name: z.string().optional(),
+ *   state: isState().optional() /// etc
+ * })
+ */
+export function isState(stateLength: 'long' | 'short' = 'short') {
+  return z.string().refine(
+    (value: any) => {
+      const input = String(value).toLowerCase()
```
RavenHursT, line 15: Should this be `any`?

Shouldn't whatever this is using be sending in a presumed `string` parameter, right?  I don't think we need runtime type-casting when TS _should_ tell us if we're linking up something from the Prisma model that we've defined as a different type, no? 🤔

RavenHursT (changes requested):

api/prisma/schema.prisma
```diff
@@ -18,6 +17,21 @@ datasource db {
...
+  id                    Int                     @id @default(autoincrement())
+  createdAt             DateTime                @default(now()) @map("created_at")
+  updatedAt             DateTime                @updatedAt @map("updated_at")
+  firstName             String                  @map("first_name")
+  lastName              String                  @map("last_name")
+  email                 String
+  phone                 String
+  metaData              Json                    @map("meta_data") @db.JsonB
+  campaignUpdateHistory CampaignUpdateHistory[]
+  campaign              Campaign[]
```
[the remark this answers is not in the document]
oldfashionedrobot: Ah good catch, should be plural. I believe it can tell the model it should be related to by the type, and the Campaign has a @relation decorator as well

api/src/shared/validations/isState.ts
oldfashionedrobot: I had refactored this function from using in a DTO, and missed this coercion. Fixed it.

api/src/campaigns/campaigns.service.ts
RavenHursT: I'm gonna go ahead and approve.. I'm fine if we just merge w/ a `TODO` comment on these raw SQL methods.. but even better if you wanna see if we can indeed build these w/ the query builder. Your call. 👍

RavenHursT (approved):

PR #11, opened 2024-11-22
Create declare module and app to app.module

oakinh:
In tgp-api this is controllers/declare/list.js


api/src/declare/declare.controller.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  @Get('list')
+  async listDeclarations() {
+    try {
+      return await this.declareService.getDeclarations();
+    } catch (error) {
+      console.error('Error at declare list:', error);
+      throw new HttpException(
+        { message: 'Error fetching declarations', error },
+        HttpStatus.BAD_REQUEST,
```
RavenHursT, line 16: If an error is being thrown out of the service, then the error would be an internal error, not a 400 Bad request, no?

api/src/declare/declare.service.ts
```diff
@@ -0,0 +1,93 @@
...
+    try {
+      const response = await fetch(url, {
+        method: 'GET',
+        headers: {
+          Authorization: `Bearer ${this.hubspotToken}`,
+        },
+      });
+
+      if (!response.ok) {
+        throw new Error(`Failed to fetch data: ${response.status} ${response.statusText}`);
```
RavenHursT, line 42: Should prolly do a `this.logger.error()` along w/ throw `BadGatewayException` here instead of a generic `Error`, then you can just   remove the `try-catch` in the controller completely since Nest will do the catching and handle it accordingly.

api/src/declare/declare.service.ts
```diff
@@ -0,0 +1,93 @@
...
+      }
+
+      const responseData = await response.json();
+
+      if (!isHubspotResponse(responseData)) {
+        throw new Error('Invalid response structure from Hubspot API');
+      }
```
RavenHursT, line 49: Log the error and throw 502 here as well

api/src/declare/declare.service.ts
```diff
@@ -0,0 +1,93 @@
...
+
+      const data = responseData.results;
+
+      const signaturesObj: Record<string, boolean> = {};
+      let signatures = '';
+
+      if (data && data.length > 0) {
```
RavenHursT, line 56: can be simplified to `data?.length` or `Boolean(data?.length)` if you want to be explicit

api/src/declare/declare.service.ts
```diff
@@ -0,0 +1,93 @@
...
+        for (const submission of data) {
+          if (submission.values.length > 0) {
+            let fn = submission.values[0].value;
+            let ln = submission.values[1].value;
```
RavenHursT, line 60: `firstName` and `lastName` here please

api/src/declare/declare.service.ts
```diff
@@ -0,0 +1,93 @@
...
+              ln =
+                ln.charAt(0).toUpperCase() + ln.slice(1).toLowerCase().trim();
+            }
+
+            const name = `${fn} ${ln}`;
+            if (!signaturesObj[name]) {
+              signatures += `${name}, `;
+              signaturesObj[name] = true;
+            }
+          }
```
RavenHursT, line 76: Could clean up all of this uniqueness-tracking w/ the two data structures and conditionals but just utilizing a `Set()` instead, correct?

api/src/declare/declare.service.ts
```diff
@@ -0,0 +1,93 @@
...
+                fn.charAt(0).toUpperCase() + fn.slice(1).toLowerCase().trim();
+            }
+            if (ln && ln.length >= 2) {
+              ln =
+                ln.charAt(0).toUpperCase() + ln.slice(1).toLowerCase().trim();
+            }
+
+            const name = `${fn} ${ln}`;
+            if (!signaturesObj[name]) {
+              signatures += `${name}, `;
```
RavenHursT, line 73: Yeah.. instead of building a string in this way, where you have to then even have the confusing conditional at the end to slice off the left over "`, `"... Just build a `Set()` and then do a `.join(', ') once it's build on the `return` and be done w/ it.

api/src/declare/declare.service.ts
```diff
@@ -0,0 +1,93 @@
...
+
+      return { signatures };
+    } catch (error) {
+      console.error('Error fetching declarations:', error);
+      throw new HttpException(
+        { message: 'Error fetching declarations', error: error.message },
+        HttpStatus.BAD_REQUEST,
+      );
+    }
+  }
```
RavenHursT, line 92: Yeah.. please remove this entire `try-catch` and just throw it around the call to hubspot so we can handle that 502 specifically instead of just throwing 400 and basically screaming into the void "something broke here!"

api/src/declare/declare.service.ts
```diff
@@ -0,0 +1,93 @@
...
+    const formId = 'f51c1352-c778-40a8-b589-b911c31e64b1';
+    const url = `https://api.hubapi.com/form-integrations/v1/submissions/forms/${formId}`;
+
+    try {
+      const response = await fetch(url, {
+        method: 'GET',
+        headers: {
+          Authorization: `Bearer ${this.hubspotToken}`,
+        },
+      });
```
RavenHursT, line 39: Can we replace all of this w/ just using their client for node? 🤔 

https://github.com/HubSpot/hubspot-api-nodejs

api/src/declare/declare.service.ts
```diff
@@ -0,0 +1,69 @@
...
+        `Failed to fetch data from HubSpot API: ${error.message}`,
+        error?.response?.data || error.stack,
+      );
+      throw new BadGatewayException('Failed to fetch data from Hubspot API');
+    }
+
+    const data = response.data?.results || [];
+
+    console.log('Full response:', response);
+    console.log('Data:', data);
```
RavenHursT, line 41: Delete dev debugging logs

api/src/declare/declare.controller.ts
```diff
@@ -0,0 +1,20 @@
...
+    try {
+      return await this.declareService.getDeclarations();
+    } catch (error) {
+      console.error('Error at declare list:', error);
+      throw new HttpException(
+        { message: 'Error fetching declarations', error },
+        HttpStatus.INTERNAL_SERVER_ERROR,
+      );
+    }
+  }
```
RavenHursT, line 19: Either delete this `try-catch`, since Nest will just automatically throw a 500 for uncaught exceptions that aren't predefined exceptions, or at least check for the `BadGatewayException` you're throwing in the service. Otherwise that `BadGatewayException` will just get swallowed here and the response will always be 500.

I think it's best to delete the `try-catch` and let uncaught exceptions bubble up to a 500, and then add other exception handling cases in the service where applicable.

RavenHursT: Also.. any logging, error or otherwise needs to be done through `this.logger.[whatever]()`, not directly through `console.log`

api/src/declare/declare.service.ts
```diff
@@ -0,0 +1,69 @@
...
+          let firstName = submission.values[0].value;
+          let lastName = submission.values[1].value;
+          // format the names to look nice and prevent duplicates.
+          if (firstName && firstName.length >= 2) {
+            firstName =
+              firstName.charAt(0).toUpperCase() + firstName.slice(1).toLowerCase().trim();
+          }
+          if (lastName && lastName.length >= 2) {
+            lastName =
+              lastName.charAt(0).toUpperCase() + lastName.slice(1).toLowerCase().trim();
```
RavenHursT, line 57: Can we move this capitalization logic to a util under `/shared/util/strings.util.ts` that looks something like:

```typescript
const capitalizeString = (s: string) => s.charAt(0).toUpperCase() + s.slice(1).toLowerCase().trim()
```

api/src/declare/declare.service.ts
```diff
@@ -0,0 +1,69 @@
...
+          }
+
+          uniqueSignatures.add(`${firstName} ${lastName}`)
+        }
+      }
+    }
+
+    const signatures = Array.from(uniqueSignatures).join(', ');
+
+    return { signatures };
```
RavenHursT, line 67: Just `return signatures` here. We're getting rid of the nested object returns that Sails forced us into w/ this new API to be more RESTful in our responses.

RavenHursT (changes requested):

RavenHursT (approved):

PR #16, opened 2024-11-27
Web 3339 races

tomer-tgp:


src/races/races.service.ts
```diff
@@ -0,0 +1,277 @@
...
+        gte: new Date(now),
+        lt: new Date(nextYear),
+      },
+    }
+    if (city && cityRecord) {
+      query['municipalityId'] = cityRecord.id
+    } else if (countyRecord) {
+      query['countyId'] = countyRecord.id
+    }
+    const races = (await this.prisma.race.findMany({
```
oldfashionedrobot, line 43: I think you could replace the `findMany` here with a `findFirst` to just return the first match

src/races/races.service.ts
```diff
@@ -0,0 +1,277 @@
...
+      countyRecord,
+      municipalityRecord,
+    )
+
+    return dedupedRaces
+  }
+
+  async byCounty(state: string, county: string) {
+    const countyRecord = await this.getCounty(state, county)
+    console.log('countyRecord', countyRecord)
```
oldfashionedrobot, line 102: console log here

RavenHursT (changes requested):
Sorry.. quite a bit here that I thought would be best addressed now rather than later. 😬 

I really do think the code and typings for this can be quite simplified. Let me know if you wanna discuss.

src/races/races.service.ts
```diff
@@ -0,0 +1,277 @@
...
+      positionSlug,
+      electionDate: {
+        gte: new Date(now),
+        lt: new Date(nextYear),
+      },
+    }
+    if (city && cityRecord) {
+      query['municipalityId'] = cityRecord.id
+    } else if (countyRecord) {
+      query['countyId'] = countyRecord.id
```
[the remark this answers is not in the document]
tomer-tgp: I got a lint error back when, but now it should be resolved. I will fix this.

src/races/races.controller.ts
```diff
@@ -0,0 +1,60 @@
...
+      const race = await this.racesService.findOne(
+        state,
+        county,
+        city,
+        positionSlug,
+      )
+      if (!race) {
+        throw new NotFoundException('Race not found')
+      }
+      return race
```
RavenHursT, line 26: Also, I noticed this later.. We really should keep the type interface on any given endpoint consistent.

That means, if an endpoint returns an `[]` of resources for a result that contains many resources, then it should _always_ return that data structure. Even if there's only 1 resource instance in the results, it's still, `[resource]`, and in the event that the filters provided result in zero results, the return ends up being 200 OK, `[]`, empty array, not 404.

404 is an HTTP status. It's saying to the client, essentially, "the address of the resource you're looking for doesn't exist", which would be the case for an endpoint like, `/races/:raceId`.  If there is no race resource that exists for the given `raceId`, then that URI, or URL, just doesn't plain exist on the internet, so that results in a 404.

But in the case of `/races?[queryFilters]`, even if the filters provided results in an `[]`, then that doesn't negate the fact that `/races` is still a valid path to a Universal Resource Identifier/Location, so that would still be a 200, and not a 404.

RavenHursT (changes requested):
One more thing I thought about last night....

RavenHursT (approved):

PR #17, opened 2024-11-28
Add TopIssue and Position tables to schema

oakinh:


tomer-tgp (changes requested):
This looks ok, but you are missing the "campaignPosition" table. That is the table where the user writes their take on the position.

so I wouldn't name position relation "campaignPosition"

TopIssue: "healthcare" (set by admins)
Position: "Free healthcare"  (set by admins)
CampaignPosition: "I want to provide free healthcare to veterans"  (set by users)

RavenHursT (approved):

prisma/schema.prisma
```diff
@@ -130,7 +130,46 @@ model Campaign {
...
   userId                Int?                    @map("user_id")
+  topIssues             TopIssue[]
+  campaignPositions     CampaignPosition[]
 
   @@index([slug])
   @@map("campaign")
 }
+
+model Position {
+  id                String             @id @default(cuid())
```
tomer-tgp, line 141: lets be consistent with the integer ids
id Int @id @default(autoincrement()) for all of these

tomer-tgp (changes requested):

prisma/schema.prisma
oakinh: Ah ok, I saw your use of it in your PR so I thought we might change everything to use cuid. I'll change it back

tomer-tgp (approved):

PR #18, opened 2024-12-03
WEB-3357: Nestjs - migrate campaign admin CRUD operations

oldfashionedrobot:
- Adds Admin Module to house admin controllers
- Adds AdminCampaigns controller + service for campaign CRUD
- Updates the way I was using Zod + ZodValidationPipe for campaigns CRUD (cleaner to use classes and not have to manually pass in the schema object to ZodValidationPipe instances)
- Updated the raw SQL query in the `campaignsService` to use Prisma query objects
- Installed `validator` package for premade validators for many cases 
- Also added a stub for the admin-create-email endpoint (for sending a password reset for an admin created campaign user)

Postman tests: https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/38094541-b98eb38b-3452-4811-947a-ba757b3ca21b

PR #19, opened 2024-12-03
[WEB-3240][WEB-3234][WEB-3235] - Authorization and User Services Implementation

RavenHursT:
https://goodparty.atlassian.net/browse/WEB-3240
https://goodparty.atlassian.net/browse/WEB-3234
https://goodparty.atlassian.net/browse/WEB-3235


src/authentication/authentication.controller.ts
```diff
@@ -0,0 +1,61 @@
...
+type LoginResult = { user: ReadUserOutputDTO; token: string }
+
+@Controller('authentication')
+export class AuthenticationController {
+  constructor(private authenticationService: AuthenticationService) {}
+
+  @Post('register')
+  @HttpCode(HttpStatus.NO_CONTENT)
+  async register(
+    @Body(new ZodValidationPipe(CreateUserInputSchema))
```
oldfashionedrobot, line 33: I found a little cleaner way to handle the `ZodValidationPipe`, instead of exporting both the zod schema object and it's inferred Type, we can just use the `createZodDto` function to make a class from the zod schema and just export that, since the class can work as a type and we also can still access the underlying zod schema from it. [An example in my PR here](https://github.com/thegoodparty/gp-api/pull/18/files#diff-9b636ce0ac854a4c0526147a0748f2746950eb77c45e422be4f6c55f477b9c3f)


Then you don't need to manually add the `ZodValidationPipe` to every handler you want to use it, you can just [add it at the top of the controller with @UsePipes()](https://github.com/thegoodparty/gp-api/pull/18/files#diff-ef7423fb013d7443572b1debd9613034809fdc675a426730ec7e262653382952) (or globally if we want to do that), and whenever the pipe sees an argument of that class it will run validations with the associated zod schema.

src/users/schemas/ReadUserOutput.schema.ts
```diff
@@ -0,0 +1,21 @@
+import { createZodDto } from 'nestjs-zod'
+import { CreateUserInputSchema } from './CreateUserInput.schema'
+import { User } from '@prisma/client'
+
+export const ReadUserOutputSchema = CreateUserInputSchema.omit({
+  password: true,
+})
+
+// TODO: Find a more explicit pattern for excluding fields
+export class ReadUserOutputDTO extends createZodDto(ReadUserOutputSchema) {
```
oldfashionedrobot, line 10: If you use the schema class approach in my comment above, you may then be able to use the `OmitType` helper method from nest like `export class ReadUserOutputSchema extends OmitType(CreateUserInputSchema, ['password']) {}` It should be type-safe so that you can't try to omit properties that don't exist

src/authentication/authentication.controller.ts
```diff
@@ -0,0 +1,61 @@
...
+  @Post('login')
+  async login(
+    @Body(new ZodValidationPipe(LoginPayloadSchema))
+    loginPayload: LoginRequestPayload,
+    @Res({ passthrough: true }) response: FastifyReply,
+  ): Promise<LoginResult> {
+    const { token, user } = await this.authenticationService.login(loginPayload)
+    setAuthToken(token, response)
+    return {
+      user: new ReadUserOutputDTO(user),
```
oldfashionedrobot, line 50: We may also just be able to bypas the DTO stuff here and just use a zod `schema.parse` to validate the output. There may be an interceptor pattern or something we could do to specify what output schema to use to parse the return value or something

src/users/schemas/ReadUserOutput.schema.ts
RavenHursT: I tried this.. but it still requires a custom `constructor` to take the resource data in and populate it. A mixture of the two approaches could possibly be warranted, but I don't think it's really worth spending that effort ATM.

Gonna resolve this and leave the TODO in place...

src/authentication/authentication.controller.ts
RavenHursT: Oh Yeeeeeah!

src/authentication/authentication.controller.ts
```diff
@@ -0,0 +1,60 @@
...
+  async register(
+    @Body() userData: CreateUserInputDto,
+    @Res({ passthrough: true }) response: FastifyReply,
+  ) {
+    const { token } = await this.authenticationService.register(userData)
+    setAuthToken(token, response)
+  }
+
+  @Post('login')
+  @UsePipes(ZodValidationPipe)
```
oldfashionedrobot, line 40: I don't think you need this `@UsePipes(ZodValidationPipe)` here since there's one on the controller level

oldfashionedrobot (commented):
All looks good to me, just found an extra `@UsePipes` decorator that I don't think you need

oldfashionedrobot (approved):

PR #23, opened 2024-12-05
Web 3236 nestjs migrate send password recovery mail

oldfashionedrobot:


src/authentication/authentication.controller.ts
```diff
@@ -56,4 +62,19 @@ export class AuthenticationController {
   async logout(@Res({ passthrough: true }) response: FastifyReply) {
     clearAuthToken(response)
   }
+
+  @Post('recover-password-email')
```
oldfashionedrobot, line 66: I put this here because that's how things were organized in `tgp-api` wasn't sure if this should be in a new controller

src/email/mailgun.service.ts
```diff
@@ -0,0 +1,31 @@
+import { Injectable } from '@nestjs/common'
+import { ConfigService } from '@nestjs/config'
+import * as FormData from 'form-data'
+import Mailgun, { MailgunMessageData } from 'mailgun.js'
+import { IMailgunClient } from 'mailgun.js/Interfaces'
+
+const EMAIL_DOMAIN = 'mg.goodparty.org'
+
+export type EmailData = MailgunMessageData & { variables?: Record<string, any> }
```
RavenHursT, line 9: Why do we need to use `any` here? Can we define this more explicitly?

src/email/mailgun.service.ts
```diff
@@ -0,0 +1,31 @@
...
+
+@Injectable()
+export class MailgunService {
+  private mailgun: Mailgun
+  private client: IMailgunClient
+
+  constructor(private config: ConfigService) {
+    this.mailgun = new Mailgun(FormData)
+    this.client = this.mailgun.client({
+      key: this.config.get('MAILGUN_API_KEY') as string,
```
RavenHursT, line 19: This is a good example of something we _should_ be able to just pull off of `process.env.MAILGUN_API_KEY`. I don't think we need this value available to anything other than this service, so probably doesn't merit the need for a specific config to this module IMHO.

src/email/util/content.util.ts
```diff
@@ -0,0 +1,298 @@
...
+GoodParty.org Team
+</div>
+  `
+}
+
+export function getBasicEmailContent(
+  msg = '',
+  // messageHeader was present in old tg-api version but not used
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  messageHeader = '',
```
RavenHursT, line 196: If this is no longer used, can we just omit it here then? What's the purpose of keeping this here w/ the comment?

RavenHursT (changes requested):

.eslintrc.js
```diff
@@ -22,5 +22,14 @@ module.exports = {
     '@typescript-eslint/explicit-function-return-type': 'off',
     '@typescript-eslint/explicit-module-boundary-types': 'off',
     '@typescript-eslint/no-explicit-any': 'off',
+    '@typescript-eslint/no-unused-vars': [
```
oldfashionedrobot, line 25: Added this to change unused vars into warnings instead of errors, and allow for prefixing with a `_` to tell TS to ignore an unused var

src/users/users.service.ts
```diff
@@ -1,33 +1,64 @@
...
-  async findUser(where: UniqueUserWhere): Promise<User | null> {
+  findUser(where: Prisma.UserWhereUniqueInput): Promise<User | null> {
     return this.prisma.user.findUnique({
       where,
     })
   }
 
+  findUserByEmail(email: string) {
+    return this.prisma.user.findFirst({
+      where: { email: { equals: email, mode: 'insensitive' } },
```
RavenHursT, line 38: Oh cool.. I like the use of the `insensitive` mode here.

But we still need to handle the `+` case as well, right?

Typically best to do this kind of normalization/sanitization of user input data on entry to the system, so that everywhere else can just assume that the data is good to act on from there on: 

https://github.com/thegoodparty/gp-api/pull/23#discussion_r1873670011

src/email/email.service.ts
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    return await this.sendEmailWithRetry(data)
+  }
+
+  async sendRecoverPasswordEmail(user: User) {
+    const { firstName, lastName, email, passwordResetToken } = user
+    const encodedEmail = email.replace('+', '%2b')
+    const link = encodeURI(
+      `${APP_BASE}/reset-password?email=${encodedEmail}&token=${passwordResetToken}`,
```
RavenHursT, line 72: If the user can be looked up by the token, then why bother encoding putting the `email` in the link?

I'm not seeing it being used anywhere here in the `/reset-password` endpoint in the `UserService` 🤔 

https://github.com/thegoodparty/gp-api/pull/23/files?diff=split&w=1#diff-090521db6f185ea4dde00b85584328375154804ef0e7b0b5a3bd4030fa91d322R49-R59

RavenHursT (changes requested):
Looks good. Just a couple of small things to wrap up.

src/email/email.service.ts
oldfashionedrobot: That url is what the front end will use to route to whatever page it needs. The email was already there in the `tgp-api` version, so didn't want to change until we get to the point where we hook up the frontend here. 

I just stubbed out the `reset-password` endpoint here just to test the validation of the JWT

RavenHursT: Ah.. ok. 

Guess we can clean it up later on then.

src/users/users.service.ts
RavenHursT: Ah.. you're correct.. I got this mixed up w/ the front-end endpoint receiving the `email` from the URL..  Mah bad.

PR #24, opened 2024-12-05
WEB-3363

oakinh:
WEB-3363
https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/collection/38601784-4eccd40b-4a17-426f-aad8-114ee48cd1f8?action=share&creator=38601784&active-environment=34024107-0205933f-8975-40bb-89f4-1c3db95f27e7


src/top-issues/top-issues.module.ts
```diff
@@ -0,0 +1,10 @@
+import { Module } from '@nestjs/common';
+import { TopIssuesController } from './top-issues.controller';
+import { TopIssuesService } from './top-issues.service';
+import { PrismaService } from 'src/prisma/prisma.service';
+
+@Module({
+  controllers: [TopIssuesController],
+  providers: [TopIssuesService, PrismaService]
```
oldfashionedrobot, line 8: If you merge in the latest dev branch, you don't need to provide the `PrismaService` here it would be provided globally

src/shared/util/svgUploader.util.ts
```diff
@@ -0,0 +1,18 @@
...
+      Key: fileName,
+      ContentType: 'image/svg+xml',
+      CacheControl: 'max-age=31536000',
+      Body: svgData,
+    },
+    bucketName,
+  );
+
+  return `https://${bucketPath}/${fileName}`;
+}
```
RavenHursT, line 18: Dead code. Please remove

src/top-issues/schemas/topIssues.schema.ts
```diff
@@ -0,0 +1,22 @@
...
+  name: z.string(), 
+  icon: z.string().nullable()
+}).strict()
+export class CreateTopIssueDto extends createZodDto(CreateTopIssueSchema) {};
+
+export const CreateTopIssueOutputSchema = z.object({
+  id: z.number(),
+  name: z.string(),
+  icon: z.string().nullable(),
+})
```
RavenHursT, line 21: Let's be more DRY w/ these schemas... 

```typescript
export const TopIssueSchema = z.object({
  id: z.number(),
  name: z.string(),
  icon: z.string().nullable(),
})
export const UpdateTopIssueSchema = z.intersection({}, TopIssueSchema) // Duplicates TopIssueSchema
export const CreateTopIssueSchema = TopIssue.omit({id: true})
```

src/top-issues/schemas/topIssues.schema.ts
```diff
@@ -0,0 +1,22 @@
+import { createZodDto } from 'nestjs-zod';
+import { z } from 'zod';
+
+export const UpdateTopIssueSchema = z.object({
+  name: z.string(),
```
RavenHursT, line 5: Where's the `id` property here? Updates should be done via `id`, no?

src/top-issues/top-issues.controller.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  @Post()
+  async createTopIssue(@Body() body: CreateTopIssueDto) {
+    return await this.topIssuesService.create(body);
+  }
+
+  @Put(':id')
+  async updateTopIssue(
+    @Param('id', ParseIntPipe) id: number,
+    @Body() body: UpdateTopIssueDto
```
RavenHursT, line 24: Oh.. I see what you did here. Yes, RESTful endpoints would have the id on the url params.. Good.

But it's perfectly acceptable to expect the ID to be on the body as well. Especially given that this is a `PUT` endpoint, and not a `PATCH` endpoint.

The RESTful expectation of `PUT` is that the client is saying to the server, "Whatever is at this URL right now, _replace_ it w/ what I'm sending you in the body, so that the next time I call a `GET` to this location, I'll get the idempotent result of exactly what I sent for the `PUT`", which would hopefully include the `id` property as well.

src/top-issues/top-issues.service.ts
```diff
@@ -0,0 +1,113 @@
...
+        this.logger.error(`Validation error ${error.message}`, error.stack);
+        throw new BadRequestException('Failed to create Top Issue - ' + error.message);
+      }
+      if (error instanceof Error) {
+        this.logger.error(`Failed to create Top Issue ${error.message}`, error.stack);
+        throw new InternalServerErrorException('An unexpected error occured.');
+      }
+      throw error;
+    }
+  }
```
RavenHursT, line 39: I say go ahead and implement that root application prisma filter, for the plethora of reasons you and I discussed and this way we don't have all these `try-catches` all over the place.

src/top-issues/top-issues.service.ts
```diff
@@ -0,0 +1,113 @@
...
+          name, 
+          icon
+        }
+      })
+
+      return {
+        id,
+        name,
+        icon
+      }
```
RavenHursT, line 27: Prisma `CRUD` operations return the resource that was acted upon:

https://www.prisma.io/docs/orm/prisma-client/queries/crud#create

Might as well use it:
```typescript
return await this.prismaService.topIssue.create({
        data: {
          name, 
          icon
        }
      })
```

src/top-issues/top-issues.service.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+  }
+
+  async update(id: number, body: UpdateTopIssueDto): Promise<TopIssue> {
+    const { name, icon } = body;
+    try {
+      return await this.prismaService.topIssue.update({
+        where: { id },
+        data: { name, icon }
+      });
```
RavenHursT, line 47: well.. you did use the resource return here 👍

src/top-issues/top-issues.service.ts
```diff
@@ -0,0 +1,113 @@
...
+      if (error instanceof PrismaClientValidationError) {
+        this.logger.error(`Validation error: ${error.message}`, error.stack);
+        throw new BadRequestException('Failed to update Top Issue - ' + error.message);
+      } else if (error instanceof Error) {
+        this.logger.error(`Failed to update Top Issue: ${error.message}`, error.stack);
+        throw new InternalServerErrorException('An unexpected error occurred.');
+      }
+      throw error;
+    }
+  }
```
RavenHursT, line 58: Abstract try-catch into the application prisma http filter handler

src/top-issues/top-issues.service.ts
```diff
@@ -0,0 +1,113 @@
...
+        await prisma.position.deleteMany({
+          where: { id: { in: positionIds } },
+        });
+      }
+
+      await prisma.topIssue.delete({
+        where: { id },
+      });
+    });
+  }
```
RavenHursT, line 96: Let's see if we can't solve this cascading deletion problem in the Prisma model definition and let the DB do it...

src/top-issues/top-issues.service.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  async list(): Promise<TopIssue[]> {
+    const topIssues = await this.prismaService.topIssue.findMany({
+      include: { positions: true }
+    })
+
+    for (let topIssue of topIssues) {
+      if (topIssue.positions && Array.isArray(topIssue.positions)) {
+        topIssue.positions.sort((a, b) => a.name.localeCompare(b.name));
+      }
```
RavenHursT, line 106: Can we let the DB and/or Prisma handle this sorting? 🤔 

https://www.prisma.io/docs/orm/prisma-client/queries/filtering-and-sorting#sorting:~:text=%5D%2C-,include%3A%20%7B,%7D%2C,-%7D)

RavenHursT (changes requested):
Also.. please add Postman collection w/ request tests and a link to a successful collection run to the description on PRs

src/top-issues/schemas/topIssues.schema.ts
oakinh: That's in the path v1/topIssues/:id

package.json
```diff
@@ -25,6 +25,7 @@
     "seed": "ts-node ./seed/seed.ts"
   },
   "dependencies": {
+    "@aws-sdk/client-s3": "^3.701.0",
```
RavenHursT, line 28: you can uninstall this since you've taken out the S3 stuff and we can add it back whenever we do that functionality

prisma/schema/position.prisma
```diff
@@ -1,7 +1,7 @@
 model Position {
   id                Int                @id @default(autoincrement())
   name              String             @unique
-  topIssue          TopIssue?          @relation(fields: [topIssueId], references: [id])
+  topIssue          TopIssue?          @relation(fields: [topIssueId], references: [id], onDelete: Cascade)
```
RavenHursT, line 4: I'm not sure we want this... if we delete a `TopIssue` we want to delete the `Position`s that relate to it.. but if we delete one position, we don't wanna delete the TopIssue associated with it, right?

src/exceptions/prisma-exception.filter.ts
```diff
@@ -0,0 +1,81 @@
...
+      message = 'Failed to intialize Prisma Client: ' + exception.message;
+
+    } else if (exception instanceof HttpException) {
+      statusCode = exception.getStatus();
+      const responseMessage = exception.getResponse();
+
+      message =
+        typeof responseMessage === 'string'
+        ? responseMessage
+        : (responseMessage as any).message || JSON.stringify(responseMessage);
```
RavenHursT, line 61: Just re-`throw` the exception again here and let NestJS handle it.

src/topIssues/topIssues.service.ts
```diff
@@ -0,0 +1,83 @@
...
+        await prisma.campaignPosition.deleteMany({
+          where: { id: { in: campaignPositionIds } },
+        });
+      }
+
+      if (positionIds.length > 0) {
+        await prisma.position.deleteMany({
+          where: { id: { in: positionIds } },
+        });
+      }
```
RavenHursT, line 60: Do we still need all of this if the table definition includes `CASCADE DELETE` now? 🤔

src/topIssues/topIssues.service.ts
```diff
@@ -0,0 +1,83 @@
...
+
+      await prisma.topIssue.delete({
+        where: { id },
+      });
+    });
+  }
+
+  async list(): Promise<TopIssue[]> {
+    const topIssues = await this.prismaService.topIssue.findMany({
```
RavenHursT, line 69: Why create the redundant variable here? Why not just `return`?

RavenHursT (changes requested):
Don't forget to add a Postman collection for these endpoints...

prisma/schema/position.prisma
oakinh: It's super weird in prisma, it works opposite what you'd expect. Notice top issue model doesn't contain any onDelete: cascade statement. This only only delete position if the top issue is deleted, not the reverse. But thinking about this made me realize we should delete campaignPositions if a position is deleted, so I added that.

https://www.prisma.io/docs/orm/prisma-schema/data-model/relations/referential-actions

oakinh:
> Also.. please add Postman collection w/ request tests and a link to a successful collection run to the description on PRs

[TopIssues.postman_test_run.json](https://github.com/user-attachments/files/18088401/TopIssues.postman_test_run.json)

RavenHursT:
> > Also.. please add Postman collection w/ request tests and a link to a successful collection run to the description on PRs
> 
> [TopIssues.postman_test_run.json](https://github.com/user-attachments/files/18088401/TopIssues.postman_test_run.json)

You don't have to upload a JSON of the run... 

Linking to the run suffices I think.. 
image

RavenHursT (approved):

PR #27, opened 2024-12-12
WEB-3237: Nestjs - migrate reset-password

oldfashionedrobot:
- Adds a `authentication/reset-password` endpoint
- Also moves the `recover-password-email` and `set-password-email` endpoints into the AuthenticationController instead of a UsersController

Postman tests: https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/38094541-3f431833-5898-472f-903e-caa10b4ff634

Postman PR: https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/pull-request/151eeff8-7eb9-4eee-849f-a73eb52bce89?workspace=66e2b59d-bf6a-4380-81c2-1cbf2d01bddf&origin=pull_request_from_fork


src/exceptions/prisma-exception.filter.ts
```diff
@@ -5,54 +5,58 @@ import {
...
-@Catch()
+const prismaErrorClasses = [
+  Prisma.PrismaClientKnownRequestError,
+  Prisma.PrismaClientUnknownRequestError,
+  Prisma.PrismaClientRustPanicError,
+  Prisma.PrismaClientInitializationError,
+  Prisma.PrismaClientValidationError,
+]
+
+@Catch(...prismaErrorClasses)
```
oldfashionedrobot, line 19: This was catching everything including Zod validation errors, updated to specifically catch prisma errors

src/users/users.controller.ts
oldfashionedrobot, line 1: Made more sense organizationally to have the auth controller manage the reset password / reset password email endpoints

package.json
```diff
@@ -22,7 +22,7 @@
     "prisma-format": "for file in prisma/schema/*.prisma; do npx prisma-case-format -c prisma/.prisma-case-format.yaml -f \"$file\"; done"
   },
   "prisma": {
-    "seed": "ts-node ./seed/seed.ts"
+    "seed": "ts-node -r tsconfig-paths/register ./seed/seed.ts"
```
RavenHursT, line 25: What is this change for? 🤔

seed/factories/user.factory.ts
```diff
@@ -16,7 +16,7 @@ export const userFactory = generateFactory<User>(() => {
     name,
     email: faker.internet.email(),
     password: generateRandomPassword(),
-    phone: faker.phone.number({ style: 'national' }),
+    phone: faker.helpers.fromRegExp(/[0-9]{10}/),
```
RavenHursT, line 19: Let's use the numeric string helper here too:
```suggestion
    phone: faker.string.numeric({ length: 10, allowLeadingZeros: false }),
```

We should prolly find a way to centralize faker field definitions like this as well.

Maybe w/ something like https://www.npmjs.com/package/@anatine/zod-mock ? 🤔

src/authentication/authentication.service.ts
```diff
@@ -44,4 +54,54 @@ export class AuthenticationService {
...
+    let user
+    try {
+      this.jwtService.verify(token)
+      user = await this.usersService.findUserByResetToken(email, token)
+    } catch (e) {
+      if (
+        e instanceof SyntaxError || // token parse failed
+        e instanceof PrismaClientKnownRequestError // token doesn't match a user
+      ) {
+        throw new BadRequestException('Invalid token')
```
RavenHursT, line 72: Should this be a `403 Forbidden` since the token (method of authorization) is invalid? 🤔

src/authentication/authentication.service.ts
```diff
@@ -44,4 +54,54 @@ export class AuthenticationService {
...
+      } else if (e instanceof TokenExpiredError) {
+        throw new BadRequestException('Token has expired')
```
RavenHursT, line 74: I think if the above is changed to `403`, then this conditional can go since either way, the token is just invalid due to exp or parsing..

src/authentication/authentication.service.ts
```diff
@@ -44,4 +54,54 @@ export class AuthenticationService {
...
+      }
+
+      throw e
+    }
+
+    try {
+      return await this.usersService.updatePassword(user.id, password, true)
+    } catch (e) {
+      console.log(e)
```
RavenHursT, line 83: Do we _need_ to catch here? If something `throw`s out of the `UserService` I would suspect we would toss the exception in such a way that we're not just defaulting to a `BadRequest`.. I mean, it could be that Prisma failed to update the record in the DB for some reason and therefor would/should be a `500 Internal` exception, right?

src/authentication/authentication.service.ts
```diff
@@ -44,4 +54,54 @@ export class AuthenticationService {
...
+      return await this.usersService.setResetToken(userId, jwt)
+    } catch (e) {
+      if (e instanceof PrismaClientKnownRequestError) {
+        console.log('Could not find user to reset password')
+        throw new NotFoundException('User not found')
+      }
+
+      throw e
+    }
+  }
```
RavenHursT, line 106: This method should just generate the token, then up in the `AuthenticationController` up where it's called, it should call `userService.update(...)` to add the token to the user record.

That way, also, the exception handling can be handled there.

Also.. we don't want to reveal to the client whether users exist or don't in our DB.. that's a security vulnerability. We can just respond w/ `204` from the controller as a way of saying, "Ok.. we accepted your request and we'll do something w/ it if it makes sense to do so"

If we want to, before responding we can do a `this.logger.warn('request to reset password for non-existent user'` if it seems valuable to do so.

RavenHursT (changes requested):

seed/factories/user.factory.ts
oldfashionedrobot: I like that package, I'll try it out with my next PR, should help simplify a lot of this

package.json
oldfashionedrobot: It's so `ts-node` can resolve relative paths in the `seed.ts` file and in files it imports. We already had that `tsconfig-paths` package installed.

src/authentication/authentication.service.ts
oldfashionedrobot: To be honest I put this catch in preemptively, so probably don't need it here. I'll handle catching any update errors in the controller level

oldfashionedrobot: And actually since that prisma exception filter will catch prisma errors, I don't need a catch just for that here anyway

RavenHursT (approved):
Cool.. just add a link to a Postman collection run to the description please:

image

src/authentication/guards/JwtAuth.guard.ts
```diff
@@ -9,11 +12,21 @@ export class JwtAuthGuard extends AuthGuard('jwt') {
...
-      ]) || super.canActivate(context)
-    )
+    const isPublic = this.reflector.getAllAndOverride(IS_PUBLIC_KEY, [
+      context.getHandler(),
+      context.getClass(),
+    ])
+
+    // If the @Roles decorator has been used, we want to override any @PublicAccess
+    // decorator that may have been applied, and require the JWT auth
+    const roles = this.reflector.getAllAndOverride<UserRole[]>(ROLES_KEY, [
```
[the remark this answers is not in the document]
RavenHursT: Oh sweet!

src/shared/util/objects.util.ts
```diff
@@ -15,7 +15,11 @@ export function deepMerge(obj1: object, obj2: object) {
 
   for (const key in obj2) {
     if (obj2.hasOwnProperty(key)) {
-      if (typeof obj2[key] === 'object' && typeof obj1[key] === 'object') {
+      if (
+        typeof obj2[key] === 'object' &&
+        typeof obj1[key] === 'object' &&
+        !Array.isArray(obj2[key])
+      ) {
```
RavenHursT, line 22: I know this whole method isn't a part of your PR.. 

But instead of reinventing the wheel here trying to maintain a custom deepmerge util,  maybe we should should replace this w/ something like https://www.npmjs.com/package/deepmerge-ts ? 🤔

seed/factories/user.factory.ts
```diff
@@ -18,6 +18,7 @@ export const userFactory = generateFactory<User>(() => {
     password: generateRandomPassword(),
     phone: faker.string.numeric({ length: 10, allowLeadingZeros: false }),
     zip: faker.location.zipCode(),
+    roles: [faker.helpers.enumValue(UserRole)],
```
RavenHursT, line 21: Is there a way to limit this to roles that are _not_ `UserRole.admin`?

I don't think it's a good idea to be seeding instances of the app w/ potentially more than 1 known, consistent admin user..

RavenHursT (changes requested):
🙏

src/shared/util/objects.util.ts
oldfashionedrobot: I think it should be fine for now, the fix here was to prevent it from trying to merge arrays. I can switch to a premade deepmerge implementation in the next PR

seed/factories/user.factory.ts
oldfashionedrobot: Updated to not use the admin role here

src/authentication/authentication.controller.ts
```diff
@@ -37,4 +52,66 @@ export class AuthenticationController {
...
+  }
+
+  @Roles(UserRole.admin)
+  @Post('set-password-email')
+  async sendSetPasswordEmail(@Body() { userId }: SetPasswordEmailSchema) {
+    const token = this.authenticationService.generatePasswordResetToken()
+    const user = await this.usersService.setResetToken(userId, token)
+    await this.emailService.sendSetPasswordEmail(user)
+
+    return { token }
```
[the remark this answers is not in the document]
oldfashionedrobot: This is the old `admin-create-email` endpoint, the `reset-password` endpoint is returning the user and token in the response. 

I've updated the naming of this to `send-set-password-email` to be more descriptive of what it actually does

src/authentication/authentication.controller.ts
```diff
@@ -37,4 +52,66 @@ export class AuthenticationController {
...
+      return
+    }
+
+    // generate and set reset token on user
+    const token = this.authenticationService.generatePasswordResetToken()
+    user = await this.usersService.setResetToken(user.id, token)
+    return await this.emailService.sendRecoverPasswordEmail(user)
+  }
+
+  @Roles(UserRole.admin)
```
RavenHursT, line 72: I think this one needs the `UserRole.sales` role to have access as well. `canCreateCampaigns` here is handling admins' and sales users' access: https://github.com/thegoodparty/tgp-api/blob/62b9ed816d49628efb3dde221479d82556befcd5/config/policies.js#L36

```suggestion
  @Roles(UserRole.admin, UserRole.sales)
```

RavenHursT (changes requested):

RavenHursT (approved):

PR #28, opened 2024-12-12
[WEB-3407] - User roles implementation

RavenHursT:
https://goodparty.atlassian.net/browse/WEB-3407

Test collection run: https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/34024107-8c31227b-55bb-4436-b435-428185ed1f92 (password recovery tests failed because those endpoints haven't been merged to `develop` as of yet)


seed/factories/user.factory.ts
```diff
@@ -14,9 +14,9 @@ export const userFactory = generateFactory<User>(() => {
     firstName,
     lastName,
     name,
-    email: faker.internet.email(),
+    email: faker.internet.email().toLowerCase(),
     password: generateRandomPassword(),
-    phone: faker.phone.number({ style: 'national' }),
+    phone: faker.number.int({ min: 1000000000, max: 9999999999 }).toString(),
```
oldfashionedrobot, line 19: Faker has a method to generate strings based on a regex, I ran into probably the same issue as you and had to use it in my PR: https://github.com/thegoodparty/gp-api/pull/27/files#diff-532e6de0c8fed1a5eef4e3998eff0373ea120abe9849a07ef9bdc5c46fb5cbe6

seed/campaigns.ts
```diff
@@ -36,6 +37,20 @@ export default async function seedCampaigns(prisma: PrismaClient) {
     fakeP2Vs[i] = p2v
   }
 
+  const ADMIN_FIRST_NAME = 'Tyler'
+  const ADMIN_LAST_NAME = 'Durden'
+  const adminUser = {
```
oldfashionedrobot, line 42: You should be able to pass this custom object into the userFactory call itself to merge those values into the generated output, something like
```suggestion
  const adminUser = userFactory({
    email: 'tyler@fightclub.org',
    password: hashSync('no1TalksAboutFightClub', genSaltSync()),
    firstName: ADMIN_FIRST_NAME,
    lastName: ADMIN_LAST_NAME,
    name: `${ADMIN_FIRST_NAME} ${ADMIN_LAST_NAME}`,
    roles: ['admin']
  })
```

src/authentication/authentication.controller.ts
```diff
@@ -1,59 +1,41 @@
...
   @Post('login')
-  async login(
-    @Body()
-    loginPayload: LoginRequestPayloadDto,
-    @Res({ passthrough: true }) response: FastifyReply,
-  ): Promise<LoginResult> {
-    const { token, user } = await this.authenticationService.login(loginPayload)
-    setAuthToken(token, response)
+  async login(@Request() { user }: RequestWithUser): Promise<LoginResult> {
+    console.log(`user =>`, user)
```
oldfashionedrobot, line 32: console.log here

src/authentication/guards/jwt-auth.guard.ts
oldfashionedrobot, line 1: I'm guessing these hyphenated filenames were generated, could we have them be camelcased for consistency? (There probably should also be a way to configure the generated casing we could look into)

oldfashionedrobot (changes requested):
All looks good to me, just found a couple cleanup issues

src/authentication/guards/jwt-auth.guard.ts
RavenHursT: Hmmm.. let me see if I can change them.. they weren't auto generated, but I was worried they might break the way Nest looks up the guards 😬 

i.e.:
https://docs.nestjs.com/recipes/passport#:~:text=class%20AppController%20%7B-,%40UseGuards(AuthGuard(%27local%27)),-%40Post(

RavenHursT: Here ya go.. worked!

https://github.com/thegoodparty/gp-api/pull/28/commits/e11032d5b75addbf2b1151c27458379d1119a0e1

oldfashionedrobot (approved):

PR #30, opened 2024-12-16
Web 3378 (part 1)

oakinh:
This is the first part of the ticket that has ~half of the transformers. Pushing this to PR now to unblock @oldfashionedrobot 
[CMS Content.postman_test_run.json](https://github.com/user-attachments/files/18153888/CMS.Content.postman_test_run.json)


package.json
```diff
@@ -66,6 +66,7 @@
     "@types/bcrypt": "^5.0.2",
     "@types/express": "^5.0.0",
     "@types/jest": "^29.5.2",
+    "@types/lodash": "^4.17.13",
```
oldfashionedrobot, line 69: We don't have `lodash` in our package.json, I think you are able to import it because it is a dependency of some other package(s) we have.

RavenHursT: Oh God. Why won't `loDash` just die already?

https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore

Please.. let's not use a massive lib just to do one small thing:

https://www.npmjs.com/package/just-camel-case

src/content/transformers/aiContentCategoriesTransformer.ts
```diff
@@ -0,0 +1,44 @@
...
+  AIContentTemplateRaw,
+  AIContentCategories
+> = (aiContents: AIContentTemplateRaw[]): AIContentCategories[] => {
+  const aiContentCategoriesHash = {};
+  const aiContentCategories: any = [];
+
+  for (const aiContent of aiContents) {
+    const { order, title } = aiContent.data.category.fields;
+    const { name } = aiContent.data;
+    const key = camelCase(name)
```
RavenHursT, line 21: https://www.npmjs.com/package/just-camel-case

src/content/transformers/candidateContentPromptsTransformer.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const logger = new Logger('CandidateContentPromptsTransformer');
+
+export const candidateContentPromptsTransformer: Transformer<
+  AIContentTemplateRaw,
+  CandidateContentPrompts
+> = (templates: AIContentTemplateRaw[]): CandidateContentPrompts[] => {
+  const result = templates.reduce<CandidateContentPrompts>((acc, template) => {
+    if (template.data.name && template.data.content) {
+      const name = camelCase(template.data.name);
```
RavenHursT, line 17: https://www.npmjs.com/package/just-camel-case

src/content/transformers/contentPromptsQuestionsTransformer.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const logger = new Logger('ContentPromptsQuestionsTransformer');
+
+export const contentPromptsQuestionsTransformer: Transformer<
+  AIContentTemplateRaw,
+  ContentPromptsQuestions
+> = (templates: AIContentTemplateRaw[]): ContentPromptsQuestions[] => {
+  const result = templates.reduce<ContentPromptsQuestions>((acc, template) => {
+    if (template.data.name && template.data.requiresAdditionalQuestions) {
+      const name = camelCase(template.data.name);
```
RavenHursT, line 17: https://www.npmjs.com/package/just-camel-case

src/content/CONTENT_TYPE_MAP.const.ts
```diff
@@ -39,10 +59,22 @@ export const CONTENT_TYPE_MAP: {
   },
   blogHome: { name: ContentType.blogHome, transformer: noOpTransformer },
   blogSection: { name: ContentType.blogSection, transformer: noOpTransformer },
+  candidateContentPrompts: {
+    name: InferredContentTypes.candidateContentPrompts,
+    transformer: candidateContentPromptsTransformer,
+    //transformer: noOpTransformer,
```
RavenHursT, line 65: delete

src/content/CONTENT_TYPE_MAP.const.ts
```diff
@@ -39,10 +59,22 @@ export const CONTENT_TYPE_MAP: {
...
+    inferredFrom: ContentType.aiContentTemplate
+  },
   candidateTestimonial: {
     name: ContentType.candidateTestimonial,
     transformer: noOpTransformer,
   },
+  contentPromptsQuestions: {
+    name: InferredContentTypes.contentPromptsQuestions,
+    transformer: contentPromptsQuestionsTransformer,
+    //transformer: noOpTransformer,
```
RavenHursT, line 75: delete

.gitignore
```diff
@@ -68,3 +68,7 @@ pids
 
 # Diagnostic reports (https://nodejs.org/api/report.html)
 report.[0-9]*.[0-9]*.[0-9]*.[0-9]*.json
+
+# just junior dev stuff
+ztestfile
```
RavenHursT, line 73: If you have a local file, in your own environment, that doesn't necessarily impact all other devs, you can always just add that file you your own local `/.git/info/exclude` file for that given repo on your own machine, then it doesn't pollute the `.gitignore` file for everyone else.

https://luisdalmolin.dev/blog/ignoring-files-in-git-without-gitignore/#:~:text=To%20ignore%20untracked%20files%2C%20you,tracking%20any%20(untracked)%20file.

src/content/transformers/aiChatPromptTransformer.ts
```diff
@@ -0,0 +1,22 @@
...
+  for (let i = 0; i < prompts.length; i++) {
+    const prompt = prompts[i];
+    const { name } = prompt.data;
+    aiChatPrompts[name] = {
+      ...prompt.data,
+      id: prompt.id
+    }
+  }
+  return [aiChatPrompts];
+}
```
RavenHursT, line 22: This could be done w/ a simple `(prompts: AIChatPrompt[]): AIChatPrompts[] => prompts.map()`, no? 🤔

src/content/transformers/aiContentCategoriesTransformer.ts
```diff
@@ -0,0 +1,44 @@
...
+  categories.sort((a, b) => a.order - b.order);
+  const combined: any = [];
+  categories.forEach((category) => {
+    combined.push({
+      name: category.title,
+      templates: categoriesHash[category.title],
+    });
+  });
+  return combined;
+}
```
RavenHursT, line 44:
```suggestion
const combineAiContentAndCategories = (categories, categoriesHash) =>
  categories.sort((a, b) => a.order - b.order).map(
    ({title}) => ({
      name: title,
      templates: categoriesHash[title],
    })
  );
```

src/content/transformers/candidateContentPromptsTransformer.ts
RavenHursT: There's also no need to create and store this in another variable if all you're doing is using it in one place on the next line.

Just do.. 

`acc[camelCase(template.data.name)]`

src/content/transformers/contentPromptsQuestionsTransformer.ts
```diff
@@ -0,0 +1,26 @@
...
+      acc[name] = template.data.requiresAdditionalQuestions;
+    } else {
+      logger.warn('template.data.name and/or template.data.requiresAdditionalQuestions not found', template);
+    }
+    return acc;
+  }, {});
+
+  return [result];
+};
```
RavenHursT, line 26: The logic here, and in `candidateContentPromptsTransformer` are essentially exactly the same, are they not?

Can we please reconcile and reuse the logic of the one for the other in some way? 🤔

src/content/transformers/onboardingPromptsTransformer.ts
```diff
@@ -0,0 +1,14 @@
...
+} from '../content.types';
+
+export const onboardingPromptsTransformer: Transformer<
+  OnboardingPromptsRaw,
+  OnboardingPromptsAugmented
+> = (prompts: OnboardingPromptsRaw[]): OnboardingPromptsAugmented[] => {
+  return prompts.map((prompt) => ({
+    ...prompt.data
+  }))
+}
```
RavenHursT, line 14: 👏 for use of `.map()` here!

FYI: no need for `return` here w/ the use of an arrow function..

src/content/transformers/promptInputFieldsTransformer.ts
```diff
@@ -0,0 +1,27 @@
...
+      const { title, helperText } = input.fields;
+      entry.push({
+        title,
+        helperText
+      })
+    }
+    promptInputFields[key] = entry;
+  }
+
+  return [promptInputFields];
```
RavenHursT, line 26: You can accomplish this logic w/ `.reduce` and `.map`

src/contentful/contentful.service.ts
```diff
@@ -25,20 +25,21 @@ const CALLS = 8
...
+  async getAllEntries() {
+    const allEntryCollections: EntryCollection<EntrySkeletonType>[] = [];
     for (let i = 0; i < CALLS; i++) {
       const entryCollection = await contentfulClient.getEntries({
         limit: LIMIT,
+        include: 5,
         skip: i * LIMIT,
-      })
-      allEntryCollections.push(entryCollection)
+      });
```
RavenHursT, line 35: You need to get `eslint` running automatically in your environment. You're sprinkling tabs, spaces, and `;`s all over the place that'll eventually break the build once we get around to enforcing it.

image

src/contentful/contentful.service.ts
```diff
@@ -25,20 +25,21 @@ const CALLS = 8
...
+      allEntryCollections.push(entryCollection);
     }
-
-    return allEntryCollections.reduce((acc, entryCollection) => {
+    const allEntries = allEntryCollections.reduce((acc, entryCollection) => {
       return [...acc, ...entryCollection.items]
     }, [] as Entry[])
+
+    return allEntries;
```
RavenHursT, line 42: Why the a redundant `allEntries` variable addition here? 🤔

src/contentful/contentful.service.ts
```diff
@@ -51,8 +52,10 @@ export class ContentfulService {
       })
     nextSyncToken = newToken as string
 
+    const allEntries = await this.getAllEntries();
+
     return {
-      allEntries: await this.getAllEntries(),
+      allEntries: allEntries,
```
RavenHursT, line 58: No need for a variable here. Just do:

```typescript
return {
  allEntries: await this.getAllEntries(),
  deletedEntries
}
```

RavenHursT (changes requested):

src/content/transformers/aiChatPromptTransformer.ts
oakinh: I don't think so, because I'm taking all of the 'aiChatPrompt' and building one object from them all, but I went ahead and changed the loop type to (const prompt of prompts) for you to make it prettier

src/content/transformers/promptInputFieldsTransformer.ts
oakinh: You're really trying to make me a FP bro huh, haha. Changed it

src/contentful/contentful.service.ts
oakinh: Was for debugging, I was printing it. I'll change it back

src/content/transformers/contentPromptsQuestionsTransformer.ts
oakinh: I looked into this, and it just seems to create an excessive abstraction that results in more code. Let me know if I'm missing something

```import { Logger } from '@nestjs/common';
import { Transformer, AIContentTemplateRaw } from '../content.types';
import camelCase from 'just-camel-case';

function createTransformer<T>(
  loggerContext: string,
  extractor: (data: AIContentTemplateRaw['data']) => [string, T] | null,
): Transformer<AIContentTemplateRaw, Record<string, T>> {
  const logger = new Logger(loggerContext);

  return (templates: AIContentTemplateRaw[]): Record<string, T>[] => {
    const result = templates.reduce<Record<string, T>>((acc, { data }) => {
      const extracted = extractor(data);
      if (extracted) {
        const [key, value] = extracted;
        acc[camelCase(key)] = value;
      } else {
        logger.warn('Required fields missing', data);
      }
      return acc;
    }, {});
    return [result];
  };
}

export const contentPromptsQuestionsTransformer = createTransformer(
  'ContentPromptsQuestionsTransformer',
  (data) =>
    data.name && data.requiresAdditionalQuestions
      ? [data.name, data.requiresAdditionalQuestions]
      : null,
);

export const candidateContentPromptsTransformer = createTransformer(
  'CandidateContentPromptsTransformer',
  (data) =>
    data.name && data.content ? [data.name, data.content] : null,
);```

RavenHursT: Ah.. I see what you're saying now.. yeah.. looks like an unhelpful exercise

src/content/transformers/promptInputFieldsTransformer.ts
RavenHursT: FYI.. it's a good habit to not be mutating the properties of function parameters, even in `.reduce`...

if you ever do end up doing a reduce, and want to have some object/array as a starting point for the `acc`.. you could end up implicitly mutating the original variable as you go through the reduce...

image

const result = arr.reduce((acc, {name}) => ({...acc, [name]: name}))

RavenHursT (approved):

PR #31, opened 2024-12-17
WEB-3358: Nestjs - migrate campaign ai CRUD operations - Part 1

oldfashionedrobot:
**NOTE: this relies on https://github.com/thegoodparty/gp-api/pull/30 - that must be merged first**

Migrates Campaign AI CRUD endpoints, up to just before enqueueing the message to kick off the AI content generation.

Postman tests: https://goodpartyorg.postman.co/workspace/66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/38094541-c8a52ad0-7e36-4776-b327-827deed9edfa

Postman PR: https://goodpartyorg.postman.co/workspace/66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/pull-request/d68dacd7-ba3b-4102-bd2f-84a3beb4d320

Required UI breaking change issues:
- https://goodparty.atlassian.net/browse/WEB-3466

### DIFF
--- a/prisma/schema/race.prisma
+++ b/prisma/schema/race.prisma
@@ -7,20 +7,22 @@ enum LevelType {
 
 model Race {
   id             Int           @id @default(autoincrement())
-  ballotId       String        @unique
-  ballotHashId   String?
-  hashId         String        @unique
-  positionSlug   String
+  ballotId       String        @unique @map("ballot_id")
+  ballotHashId   String?       @map("ballot_hash_id")
+  hashId         String        @unique @map("hash_id")
+  positionSlug   String        @map("position_slug")
   state          String        @db.Char(2)
-  electionDate   DateTime?
+  electionDate   DateTime?     @map("election_date")
   level          LevelType
-  subAreaName    String
-  subAreaValue   String
+  subAreaName    String        @map("sub_area_name")
+  subAreaValue   String        @map("sub_area_value")
   data           Json          @default("{}") @db.JsonB
   county         County?       @relation(fields: [countyId], references: [id])
-  countyId       Int?
+  countyId       Int?          @map("county_id")
   municipality   Municipality? @relation(fields: [municipalityId], references: [id])
-  municipalityId Int?
+  municipalityId Int?          @map("municipality_id")
+
+  @@map("race")
 }
 
 model County {
@@ -31,6 +33,8 @@ model County {
   data           Json?          @default("{}") @db.JsonB
   municipalities Municipality[]
   races          Race[]
+
+  @@map("county")
 }
 
 enum MunicipalityType {
@@ -48,6 +52,8 @@ model Municipality {
   state    String           @db.Char(2)
   data     Json?            @default("{}") @db.JsonB
   county   County?          @relation(fields: [countyId], references: [id])
-  countyId Int?
+  countyId Int?             @map("county_id")
   races    Race[]
+
+  @@map("municipality")
 }
--- a/src/campaigns/ai/campaignsAi.controller.ts
+++ b/src/campaigns/ai/campaignsAi.controller.ts
@@ -30,11 +30,10 @@ export class CampaignsAiController {
   @Post()
   async create(
     @Res({ passthrough: true }) res: FastifyReply,
-    @ReqUser() user: User,
+    @ReqUser() { id: userId }: User,
     @Body() body: CreateAiContentSchema,
   ) {
-    // TODO: use a decorator to inject needed campaign for user instead of this findByUser everywhere
-    const campaign = await this.campaignsService.findByUser(user.id)
+    const campaign = await this.loadCampaign(userId)
     const result = await this.campaignsAiService.createContent(campaign, body)
 
     if (result.created) {
@@ -48,15 +47,23 @@ export class CampaignsAiController {
 
   @Put('rename')
   @HttpCode(HttpStatus.OK)
-  async rename(@ReqUser() user: User, @Body() body: RenameAiContentSchema) {
-    const campaign = await this.campaignsService.findByUser(user.id)
+  async rename(
+    @ReqUser() { id: userId }: User,
+    @Body() body: RenameAiContentSchema,
+  ) {
+    const campaign = await this.loadCampaign(userId)
     return this.campaignsAiService.updateContentName(campaign, body)
   }
 
   @Delete(':key')
   @HttpCode(HttpStatus.NO_CONTENT)
-  async delete(@ReqUser() user: User, @Param('key') key: string) {
-    const campaign = await this.campaignsService.findByUser(user.id)
+  async delete(@ReqUser() { id: userId }: User, @Param('key') key: string) {
+    const campaign = await this.loadCampaign(userId)
     return this.campaignsAiService.deleteContent(campaign, key)
+  }
+
+  private loadCampaign(userId: number) {
+    // TODO: use a decorator to inject needed campaign for user instead of this findByUser everywhere
+    return this.campaignsService.findByUser(userId)
   }
 }
--- a/src/campaigns/ai/campaignsAi.controller.ts
+++ b/src/campaigns/ai/campaignsAi.controller.ts
@@ -31,7 +31,7 @@ export class CampaignsAiController {
   ) {
     const result = await this.aiService.createContent(user.id, body)
 
-    if (result.step === 'created') {
+    if (result.created) {
       res.statusCode = HttpStatus.CREATED
     } else {
       res.statusCode = HttpStatus.OK
--- a/src/campaigns/ai/campaignsAi.controller.ts
+++ b/src/campaigns/ai/campaignsAi.controller.ts
@@ -2,6 +2,7 @@ import {
   Body,
   Controller,
   Delete,
+  HttpCode,
   HttpStatus,
   Param,
   Post,
@@ -19,7 +20,6 @@ import { FastifyReply } from 'fastify'
 
 @Controller('campaigns/ai')
 @UsePipes(ZodValidationPipe)
-// @UseGuards(CampaignOwnersOrAdminGuard) // TODO: need guard to check user has campaign?
 export class CampaignsAiController {
   constructor(private aiService: CampaignsAiService) {}
 
@@ -40,7 +40,8 @@ export class CampaignsAiController {
     return result
   }
 
-  @Post('rename')
+  @Post('rename') // TODO: should be a PATCH instead?
+  @HttpCode(HttpStatus.OK)
   rename(@ReqUser() user: User, @Body() body: RenameAiContentSchema) {
     return this.aiService.updateContentName(user.id, body)
   }
--- a/src/campaigns/ai/campaignsAi.controller.ts
+++ b/src/campaigns/ai/campaignsAi.controller.ts
@@ -2,7 +2,10 @@ import {
   Body,
   Controller,
   Delete,
+  HttpStatus,
+  Param,
   Post,
+  Req,
   UseGuards,
   UsePipes,
 } from '@nestjs/common'
@@ -11,8 +14,8 @@ import { ReqUser } from 'src/authentication/decorators/ReqUser.decorator'
 import { RenameAiContentSchema } from './schemas/RenameAiContent.schema'
 import { ZodValidationPipe } from 'nestjs-zod'
 import { User } from '@prisma/client'
-import { DeleteAiContentSchema } from './schemas/DeleteAiContent.schema'
 import { CreateAiContentSchema } from './schemas/CreateAiContent.schema'
+import { FastifyReply } from 'fastify'
 
 @Controller('campaigns/ai')
 @UsePipes(ZodValidationPipe)
@@ -21,8 +24,20 @@ export class CampaignsAiController {
   constructor(private aiService: CampaignsAiService) {}
 
   @Post()
-  create(@ReqUser() user: User, @Body() body: CreateAiContentSchema) {
-    return this.aiService.createContent(user.id, body)
+  async create(
+    @Req() req: FastifyReply,
+    @ReqUser() user: User,
+    @Body() body: CreateAiContentSchema,
+  ) {
+    const result = await this.aiService.createContent(user.id, body)
+
+    if (result.step === 'created') {
+      req.statusCode = HttpStatus.CREATED
+    } else {
+      req.statusCode = HttpStatus.OK
+    }
+
+    return result
   }
 
   @Post('rename')
@@ -30,8 +45,8 @@ export class CampaignsAiController {
     return this.aiService.updateContentName(user.id, body)
   }
 
-  @Delete()
-  delete(@ReqUser() user: User, @Body() { key }: DeleteAiContentSchema) {
+  @Delete(':key')
+  delete(@ReqUser() user: User, @Param('key') key: string) {
     return this.aiService.deleteContent(user.id, key)
   }
 }
--- a/src/campaigns/ai/campaignsAi.controller.ts
+++ b/src/campaigns/ai/campaignsAi.controller.ts
@@ -6,8 +6,7 @@ import {
   HttpStatus,
   Param,
   Post,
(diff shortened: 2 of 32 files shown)


prisma/schema/race.prisma
```diff
@@ -7,20 +7,22 @@ enum LevelType {
 
 model Race {
   id             Int           @id @default(autoincrement())
-  ballotId       String        @unique
-  ballotHashId   String?
-  hashId         String        @unique
-  positionSlug   String
+  ballotId       String        @unique @map("ballot_id")
```
oldfashionedrobot, line 10: Updated casing here to match our downcased / snakecased pattern

src/campaigns/ai/campaignsAi.service.ts
```diff
@@ -0,0 +1,534 @@
...
+import { positionsToStr, againstToStr, replaceAll } from './util/prompt.util'
+
+@Injectable()
+export class CampaignsAiService {
+  constructor(
+    private campaignsService: CampaignsService,
+    private contentService: ContentService,
+  ) {}
+
+  async createContent(userId: number, inputs: CreateAiContentSchema) {
```
oldfashionedrobot, line 25: Migrates: https://github.com/thegoodparty/tgp-api/blob/develop/api/controllers/campaign/ai/create.js

src/campaigns/ai/campaignsAi.service.ts
```diff
@@ -0,0 +1,534 @@
...
+    // await sails.helpers.queue.consumer()
+
+    return {
+      status: 'processing',
+      step: 'created',
+      key,
+    }
+  }
+
+  async updateContentName(userId: number, inputs: RenameAiContentSchema) {
```
oldfashionedrobot, line 151: Migrates: https://github.com/thegoodparty/tgp-api/blob/develop/api/controllers/campaign/ai/rename.js

src/campaigns/ai/campaignsAi.service.ts
```diff
@@ -0,0 +1,534 @@
...
+    }
+
+    aiContent[key]['name'] = name
+
+    return this.campaignsService.update(campaign.id, {
+      aiContent,
+    })
+  }
+
+  async deleteContent(userId: number, aiContentKey: string) {
```
oldfashionedrobot, line 168: Migrates: https://github.com/thegoodparty/tgp-api/blob/develop/api/controllers/campaign/ai/delete.js

src/campaigns/ai/util/prompt.util.ts
```diff
@@ -0,0 +1,55 @@
+import { Prisma } from '@prisma/client'
+import { CampaignDetailsContent } from 'src/campaigns/campaigns.types'
+
+type CampaignPositionWithRelations = Prisma.CampaignPositionGetPayload<{
+  include: { topIssue: true; position: true }
+}>
+
+export function positionsToStr(
```
oldfashionedrobot, line 8: Utility functions used in `promptReplace` method, migrated from: https://github.com/thegoodparty/tgp-api/blob/develop/api/helpers/ai/prompt-replace.js

src/content/content.service.ts
```diff
@@ -57,6 +61,35 @@ export class ContentService {
     return transformContent(ContentType.glossaryItem, entries)
   }
 
+  async getAiContentPrompts() {
```
oldfashionedrobot, line 64: Migrates: https://github.com/thegoodparty/tgp-api/blob/develop/api/helpers/ai/get-prompts.js

src/campaigns/ai/util/prompt.util.ts
```diff
@@ -0,0 +1,55 @@
...
+      str += `${issue?.title} - ${issue?.position}, `
+    })
+  }
+  return str
+}
+
+export function replaceAll(string: string, search: string, replace: string) {
+  const replaceStr = replace || 'unknown'
+  return string.split(`[[${search}]]`).join(replaceStr)
+}
```
RavenHursT, line 36: Is there a reason why we're reinventing `String.prototype.replaceAll` here? 🤔 

I mean.. this could be replaced w/ 

`stringToModify.replaceAll(`[[${search}]]`, replace || 'unknown')` right?

src/campaigns/ai/campaignsAi.service.ts
```diff
@@ -0,0 +1,534 @@
...
+        try {
+          newPrompt = replaceAll(
+            newPrompt,
+            item.find,
+            item.replace ? item.replace.toString().trim() : '',
+          )
+        } catch (e) {
+          console.log('error at prompt replace', e)
+        }
+      })
```
RavenHursT, line 521: this could totally be replaced w/ a `.reduce()` using native `.replaceAll()`, right?

src/campaigns/ai/campaignsAi.service.ts
```diff
@@ -0,0 +1,534 @@
...
+      `
+
+      return newPrompt
+    } catch (e) {
+      console.log('Error in helpers/ai/promptReplace', e)
+      //TODO: surface relevant error here
+      return ''
+    }
+  }
+}
```
RavenHursT, line 534: this method _really_ needs to be broken down into the relevant logic pieces regarding how getting a new prompt, produces side-effects in, and depends on, `pathToVictory, campaignPositions, campaignUpdateHistory` and the `user` associated w/ the given campaign.

Seems like, w/ a little thought this could be greatly simplified and would allow each component of the logic flow here to be separated responsibly.

src/campaigns/campaigns.controller.ts
```diff
@@ -73,7 +73,7 @@ export class CampaignsController {
     @Body() body: UpdateCampaignSchema,
   ) {
     // TODO get campaign from req user
-    const updateResp = await this.campaignsService.update(id, body)
+    const updateResp = await this.campaignsService.updateJsonFields(id, body)
```
RavenHursT, line 76: If this method is only updating those few JSONb fields on the `campaign` table, then how do we update the other columns for a given `campaign`? Can we not reconcile this functionality into a singular `update` method on the service? 🤔 

Looks like we may be able to even branch this update logic based on the type of the Prisma model fields being updated:
Screenshot 2024-12-17 at 4 42 46 PM

src/content/content.service.ts
```diff
@@ -57,6 +61,35 @@ export class ContentService {
...
+          {
+            type: ContentType.candidateContentPrompts,
+          },
+        ],
+      },
+    })
+
+    // should be one content record for each "type" of prompt
+    if (prompts.length !== 2) {
+      throw new NotFoundException('Prompt content not found')
```
RavenHursT, line 80: This should be a `BadGatewayException` since this has to do w/ our integration with Contentful, not something the requesting service is asking for.

RavenHursT: I would also do a check for `.data` on both prompts here, and that way the try-catch below can be removed.

src/content/content.service.ts
```diff
@@ -57,6 +61,35 @@ export class ContentService {
...
+    }
+
+    try {
+      return {
+        ...(prompts[0].data as Prisma.JsonObject),
+        ...(prompts[1].data as Prisma.JsonObject),
+      }
+    } catch (_e) {
+      throw new InternalServerErrorException('Failed to load prompts')
+    }
```
RavenHursT, line 90: This is just a try wrapping around a `return`. The only thing that could possibly `throw` here would be trying to spread on `.data`, if for some reason `.data` wasn't returned from Contentful for some reason. In that case, it should be handled as a `BadGatewayException` as well in the `if` block above.

src/users/util/users.util.ts
```diff
@@ -0,0 +1,5 @@
+import { User } from '@prisma/client'
+
+export function getFullName(user: User) {
```
RavenHursT, line 3: ❤️ 👏

RavenHursT (changes requested):

src/campaigns/ai/campaignsAi.service.ts
```diff
@@ -0,0 +1,534 @@
...
+      status: 'processing',
+      step: 'created',
+      key,
+    }
+  }
+
+  async updateContentName(userId: number, inputs: RenameAiContentSchema) {
+    const { key, name } = inputs
+
+    const campaign = await this.campaignsService.findByUser(userId)
```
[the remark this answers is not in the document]
oldfashionedrobot: That's a good idea, I'll make the change

src/campaigns/ai/campaignsAi.service.ts
```diff
@@ -0,0 +1,534 @@
...
+    const campaign = await this.campaignsService.findByUser(userId, {
+      pathToVictory: true,
+      campaignPositions: {
+        include: {
+          topIssue: true,
+          position: true,
+        },
+      },
+      campaignUpdateHistory: true,
+      user: true,
```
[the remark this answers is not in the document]
oldfashionedrobot: The rest of these are used in the promptReplace method called later, but I can separate it to just the data needed up here, and make a separate query for the rest if it makes it to the promptReplace method

src/campaigns/ai/campaignsAi.service.ts
```diff
@@ -0,0 +1,534 @@
...
+    const keyNoDigits = key.replace(/\d+/g, '') // we allow multiple keys like key1, key2
+    let prompt = cmsPrompts[keyNoDigits] as string
+
+    prompt = await this.promptReplace(prompt, campaign)
+    if (!prompt || prompt === '') {
+      // await sails.helpers.slack.errorLoggerHelper('empty prompt replace', {
+      //   cmsPrompt: cmsPrompts[keyNoDigits],
+      //   promptAfterReplace: prompt,
+      //   campaignObj,
+      // })
```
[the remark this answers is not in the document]
oldfashionedrobot: There are a lot of logging + slack logging that is in these methods and in the Part 2 PR. I made a followup ticket to go through and update those places to use the logger: https://goodparty.atlassian.net/browse/WEB-3462

src/campaigns/ai/campaignsAi.service.ts
```diff
@@ -0,0 +1,534 @@
...
+    // })
+
+    if (!aiContent.generationStatus[key]) {
+      aiContent.generationStatus[key] = {} as AiContentGenerationStatus
+    }
+    aiContent.generationStatus[key].status = 'processing'
+    aiContent.generationStatus[key].prompt = prompt as string
+    aiContent.generationStatus[key].existingChat = chat || []
+    aiContent.generationStatus[key].inputValues = inputValues
+    aiContent.generationStatus[key].createdAt = new Date().valueOf()
```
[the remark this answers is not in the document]
oldfashionedrobot: I'm not sure if this reads better than just setting the properties individually here. Also, the intent of this PR was to migrate the code/functions over without changing too much around. I don't completely know the context of all this code. 

I made a follow up ticket that I can do some code cleanup in after I can make sure the functionality works: https://goodparty.atlassian.net/browse/WEB-3462

src/campaigns/ai/util/prompt.util.ts
oldfashionedrobot: from another comment: https://github.com/thegoodparty/gp-api/pull/31#discussion_r1890409319
> the intent of this PR was to migrate the code/functions over without changing too much around. I don't completely know the context of all this code.
> 
>  made a follow up ticket that I can do some code cleanup in after I can make sure the functionality works: https://goodparty.atlassian.net/browse/WEB-3462

src/campaigns/ai/campaignsAi.service.ts
oldfashionedrobot: same comment as above

src/campaigns/ai/campaignsAi.service.ts
oldfashionedrobot: same comment as above, can deal with cleaning this up in followup

src/campaigns/ai/campaignsAi.service.ts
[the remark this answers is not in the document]
oldfashionedrobot: Actually on second thought, I think method should stay how it is, it is a super specific method to just change the name of an AI content JSON object. 

It's only really extracted into this service method to clean up the controller. If we get to a point where we need a more general "update Ai content" method we could change this then

src/campaigns/ai/campaignsAi.controller.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    return result
+  }
+
+  @Post('rename') // TODO: should be a PATCH instead?
+  @HttpCode(HttpStatus.OK)
+  rename(@ReqUser() user: User, @Body() body: RenameAiContentSchema) {
+    return this.aiService.updateContentName(user.id, body)
+  }
```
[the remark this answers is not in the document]
oldfashionedrobot: https://goodparty.atlassian.net/browse/WEB-3466

src/campaigns/ai/campaignsAi.service.ts
```diff
@@ -0,0 +1,534 @@
...
+
+    // await sails.helpers.queue.enqueue(queueMessage)
+    // await sails.helpers.slack.aiLoggerHelper('Enqueued AI prompt', queueMessage)
+
+    // why do we call the consumer twice ?
+    // await sails.helpers.queue.consumer()
+
+    return {
+      status: 'processing',
+      step: 'created',
```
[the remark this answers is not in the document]
oldfashionedrobot: I don't think that step is being used by the frontend currently, so I will remove that key from the response

src/campaigns/ai/campaignsAi.service.ts
```diff
@@ -0,0 +1,534 @@
...
+    if (!aiContent?.[key]) {
+      throw new BadRequestException('Invalid document key')
+    }
+
+    aiContent[key]['name'] = name
+
+    return this.campaignsService.update(campaign.id, {
+      aiContent,
+    })
+  }
```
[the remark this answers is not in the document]
oldfashionedrobot: See comment here: https://github.com/thegoodparty/gp-api/pull/31#discussion_r1890521214

src/content/content.service.ts
oldfashionedrobot: Updated to check for data instead of the try/catch. But I'm not sure this would be a BadGatewayException, since it is just querying our DB.

src/campaigns/ai/campaignsAi.service.ts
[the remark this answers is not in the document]
RavenHursT: We can make the generic `updateCampaign` later on, cool.

But let's go ahead and move the `.findByUser()` up into a single method on the controller w/ a `TODO: Make this a decorator` comment, then we can just pass down `aiContent` and the `contentId` into these service methods to be more DRY and decouple the user data from this service for `aiContent` operations.

src/content/content.service.ts
RavenHursT: It's data effectively cached from a third-party service. If we weren't storing that data in our DB, and just calling Contentful directly, we'd be returning `BadGatewayException` here, not a 404. 

This method is also only called by the `CampaignsAiService` service, so it's not being called by the client.  A 404, bubbling up through the `CampaignsAiService` would give the impression that the `aiContent` for the given Campaign isn't there, when that's not what's happening there, the _Content_ for to generate new `aiContent` are missing/corrupted.  That's a 502 for that integration.

RavenHursT (changes requested):
Ok.. just a couple of remaining things.. (I think you already addressed the first one)

- https://github.com/thegoodparty/gp-api/pull/31#discussion_r1890851765
- https://github.com/thegoodparty/gp-api/pull/31#discussion_r1890899522

src/content/content.service.ts
oldfashionedrobot: Now that you mention it, this probably shouldn't be throwing HTTP exceptions. I updated to just throw a regular error, and if a controller needs to catch and throw an HTTP error it can

PR #33, opened 2024-12-18
feat: WEB-3409

taylorgp:


src/shared/util/graphql.util.ts
```diff
@@ -0,0 +1,28 @@
...
+      ${query}
+    `
+
+    const data: any = await graphQLClient.request(gqlQuery, variables || {})
+    return data
+  } catch (e) {
+    logger.error('error at queryHelper', e)
+    throw e
+  }
+}
```
RavenHursT, line 28: Pretty sure this should be a service w/ the `GraphQLClient` as a `private` singleton on the service class, instead of instantiating a new instance w/ each and every request.

The `endpoint` can either be passed into the service, if we're going to be using this as an interface to talk to multiple different GraphQL services, or it can just be a standalone constant in this file instead of creating a new string literal for each request as well.

src/races/util/getRace.util.ts
```diff
@@ -0,0 +1,48 @@
...
+                  }
+                  filingPeriods {
+                      endOn
+                      startOn
+                  }
+              }
+          }
+      }
+      `
+  const { node } = await queryHelper(query)
```
RavenHursT, line 46: If you want to keep these query string builders in util methods, that'll make sense. But the GQL service request should be called from the `RacesService`, not in a util method.

src/races/util/getElectionDates.util.ts
```diff
@@ -0,0 +1,60 @@
...
+      }
+    }
+    logger.log(slug, 'electionDates', electionDates)
+
+    return electionDates
+  } catch (e) {
+    logger.error(slug, 'error at extract-location-ai helper', e)
+    return []
+  }
+}
```
RavenHursT, line 60: This is doing a lot more than just building a GQL query. Seems like a good candidate for a method in `RacesService`.  If you want to pull out the query building into a util, that's fine, but there's a good bit of business logic in there that would be better placed in a service method.

src/races/races.service.ts
```diff
@@ -294,4 +296,534 @@ export class RacesService {
...
+        level !== 'county' &&
+        level !== 'city'
+      ) {
+        level = 'city'
+      }
+      return level
+    } catch (e) {
+      this.logger.log('error at get-race-level', e)
+      return level
+    }
```
RavenHursT, line 330: Not sure we need this `try-catch` right?  The only thing that could `throw` here is if `level` ends up not a `String` somehow.. and that shouldn't happen since we've declared it as such in the method args.

Even if we did want to protect against a runtime possibility there, we'd wrap just the `.toLowerCase()` here, not the entire conditional and `return`.

prisma/schema/race.prisma
```diff
@@ -51,3 +51,12 @@ model Municipality {
...
 }
+
+model CensusEntity {
+  id       Int      @id @default(autoincrement())
+  mtfcc    String
+  mtfccType String
+  geoId    String
+  name     String
+  state    String
+}
```
RavenHursT, line 62: Needs `@map` and `@@map` for snake cases

src/races/races.service.ts
```diff
@@ -294,4 +296,534 @@ export class RacesService {
...
+  }
+
+  private async resolveMtfcc(geoId: string, mtfcc: string) {
+    let geoData: any
```
RavenHursT, line 334: Can we please define this? Doesn't look like it has _all that many_ properties.

src/races/races.service.ts
```diff
@@ -294,4 +296,534 @@ export class RacesService {
...
+          name: census.name,
+          type: census.mtfccType,
+        }
+
+        // todo: this can be improved for county recognition
+        // and other types of entities (school board, etc)
+
+        // if (census.mtfccType !== 'State or Equivalent Feature') {
+        //   geoData.city = census.name;
+        // }
```
RavenHursT, line 357: delete commented code.

src/races/races.service.ts
```diff
@@ -294,4 +296,534 @@ export class RacesService {
...
+            geoData.township = census.name
+          } else if (census.name.toLowerCase().includes('town')) {
+            geoData.town = census.name
+          } else if (census.name.toLowerCase().includes('city')) {
+            geoData.city = census.name
+          } else if (census.name.toLowerCase().includes('village')) {
+            geoData.village = census.name
+          } else if (census.name.toLowerCase().includes('borough')) {
+            geoData.borough = census.name
+          }
```
RavenHursT, line 376: All of these string literals need to be placed into `enum`s and referenced as such

src/races/races.service.ts
```diff
@@ -294,4 +296,534 @@ export class RacesService {
...
+    try {
+      electionDate = race?.election?.electionDay
+      termLength = race?.position?.electionFrequencies[0].frequency[0]
+      level = race?.position?.level.toLowerCase()
+      positionId = race?.position.id
+      mtfcc = race?.position.mtfcc
+      geoId = race?.position.geoId
+      tier = race?.position.tier
+    } catch (e) {
+      logger.log(slug, 'error getting election date', e)
```
RavenHursT, line 420: There's only assignments happening here. We don't need this `try-catch`.

src/races/races.service.ts
```diff
@@ -294,4 +296,534 @@ export class RacesService {
...
+    }
+    if (!electionDate) {
+      return
+    }
+
+    let electionLevel = 'city'
+    try {
+      electionLevel = this.getRaceLevel(level)
+    } catch (e) {
+      logger.log(slug, 'error getting election level', e)
```
RavenHursT, line 430: Should this not be `logger.error`?

src/races/races.service.ts
RavenHursT, line 420: `logger.error`?

src/races/races.service.ts
```diff
@@ -294,4 +296,534 @@ export class RacesService {
...
+    const logger = new Logger('getRaceDetails')
+    const data: any = {}
+
+    logger.log(slug, 'getting race from ballotReady api...')
+
+    let race: any
+    try {
+      race = await getRaceById(raceId)
+    } catch (e) {
+      logger.log(slug, 'error getting race details', e)
```
RavenHursT, line 398: `logger.error`?

src/races/races.service.ts
```diff
@@ -294,4 +296,534 @@ export class RacesService {
...
+    try {
+      electionLevel = this.getRaceLevel(level)
+    } catch (e) {
+      logger.log(slug, 'error getting election level', e)
+    }
+    logger.log(slug, 'electionLevel', electionLevel)
+
+    const officeName = race?.position?.name
+    if (!officeName) {
+      logger.log(slug, 'error getting office name')
```
RavenHursT, line 436: `logger.error`?

src/races/races.service.ts
```diff
@@ -294,4 +296,534 @@ export class RacesService {
...
+    data.electionState = electionState
+    data.electionCounty = county
+    data.electionMunicipality = city
+    data.subAreaName = subAreaName
+    data.subAreaValue = subAreaValue
+    data.partisanType = partisanType
+    data.priorElectionDates = priorElectionDates
+    data.positionId = positionId
+    data.tier = tier
+    return data
```
RavenHursT, line 597: Instead of having this in one massive method, we _could_ split these out into separate `private` service methods, or even module-level util methods, depending on complexity and dependency requirements.

If we don't want to do that right now, can we at least add a `// TODO:` here to do just that?

src/races/races.service.ts
```diff
@@ -294,4 +296,534 @@ export class RacesService {
...
+        decodedContent.level = level
+      } catch (e) {
+        console.log('error at extract-location-ai helper', e)
+      }
+      console.log('extract ai location response', decodedContent)
+      return decodedContent
+    } catch (e) {
+      console.log('error at extract-location-ai helper', e)
+      return false
+    }
```
RavenHursT, line 827: Can we please try to target our `try-catch`es?

There's a lot of just native JS functionality here, no service calls, unless I'm missing something... 🤔 

Everything here looks like it can just be handled by either existential checks, or, in the case of the `JSON.parse()`, `try-catch`ing _just_ around that one operation to protect against a non-JSON string somehow getting through.

src/races/races.service.ts
```diff
@@ -294,4 +296,534 @@ export class RacesService {
...
+      }
+      Example Input: "US Senate - CA"
+      Example Output: {
+      }
+      Example Input: "Maine Village Board Chair - Wisconsin"
+      Example Output:
+      {
+           "village": "Maine Village",
+           "county": "Marathon County"
+      }
```
RavenHursT, line 731: These can be moved out to a `prompts.consts.ts` file

src/races/races.service.ts
```diff
@@ -294,4 +296,534 @@ export class RacesService {
...
+      try {
+        decodedContent = JSON.parse(content)
+        decodedContent.level = level
+      } catch (e) {
+        console.log('error at extract-location-ai helper', e)
+      }
+      console.log('extract ai location response', decodedContent)
+      return decodedContent
+    } catch (e) {
+      console.log('error at extract-location-ai helper', e)
```
RavenHursT, line 825: please use `this.logger` and the appropriate level method for logging

RavenHursT (changes requested):

src/graphql/graphql.service.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  async fetchGraphql(query: any, variables?: any) {
+    let data: any
+
+    const graphQLClient = new GraphQLClient(API_BASE, {
+      headers: {
+        authorization: `Bearer ${BALLOT_READY_KEY}`,
+        'Content-Type': 'application/json',
+      },
+    })
```
RavenHursT, line 21: This is still being created on _every_ call to this method.

This should be a `private` member on this class as mentioned in the last PR here: https://github.com/thegoodparty/gp-api/pull/33#discussion_r1890938951

prisma/schema/race.prisma
```diff
@@ -57,3 +57,14 @@ model Municipality {
...
+
+model CensusEntity {
+  id       Int      @id @default(autoincrement())
+  mtfcc    String
+  mtfccType String
+  geoId    String
+  name     String
+  state    String
+
+  @@map("censusEntity")
```
oldfashionedrobot, line 69: We've been snakecasing all the camelcased table names and field names. You could run `npm run prisma-format` to automatically add the relevant `@map`s here

src/graphql/graphql.module.ts
```diff
@@ -0,0 +1,10 @@
+import { Module } from '@nestjs/common'
+import { GraphqlService } from './graphql.service'
+import { GraphQLClient } from 'graphql-request'
+
+@Module({
+  controllers: [],
+  providers: [GraphqlService],
+  imports: [GraphQLClient],
```
RavenHursT, line 8: I don't think you need this line, `imports` are for when you're trying to utilize providers from other modules in nestJS: https://stackoverflow.com/a/73751722/282250

src/races/races.service.ts
RavenHursT: This still isn't addressed.

src/races/races.service.ts
RavenHursT: This still isn't addressed.

src/races/races.service.ts
RavenHursT: These still aren't addressed

oldfashionedrobot (changes requested):

src/races/races.service.ts
RavenHursT: I'm not seeing this was addressed?

src/races/races.service.ts
```diff
@@ -294,4 +305,521 @@ export class RacesService {
...
+                            election {
+                                electionDay
+                            }
+                        }
+                    }
+                }
+            }`
+
+      const { races } = await this.graphQLService.fetchGraphql(query)
+      logger.log(slug, 'getElectionDates graphql result', races)
```
RavenHursT, line 757: Why not use `this.logger` in this method? Why create another one?

src/races/races.service.ts
RavenHursT: This isn't addressed

PR #36, opened 2024-12-20
[WEB-3420] - Google OAuth2 Implementation

RavenHursT:
https://goodparty.atlassian.net/browse/WEB-3420

Postman test run: https://goodpartyorg.postman.co/workspace/66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/34024107-01d6afb9-4691-4a06-b2b3-77e7624b3f4b

Regression tests run of username/password auth: https://goodpartyorg.postman.co/workspace/66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/34024107-b5b26a3d-c32b-430c-ab76-9734c2147747


src/authentication/authentication.controller.ts
```diff
@@ -57,6 +58,20 @@ export class AuthenticationController {
     }
   }
 
+  @UseGuards(AuthGuard(SOCIAL_LOGIN_STRATEGY_NAME))
+  @Post('social-login/:socialProvider')
+  async socialLogin(
+    @Request() { user }: RequestWithUser,
```
oldfashionedrobot, line 64: Should this use the `@ReqUser` decorator?

src/users/users.controller.ts
```diff
@@ -48,6 +48,14 @@ export class UsersController {
     return ReadUserOutputSchema.parse(dbUser)
   }
 
+  @UseGuards(UserOwnerOrAdminGuard)
```
oldfashionedrobot, line 51: Since there is no `id` param here, this guard is really just checking for an admin role right? Could maybe just use the `@Roles` decorator?

RavenHursT: If I do `@Roles(UserRoles.admin)` then any user that tries to request `/users/me`, who doesn't have the `admin` role, will be denied access.

I know the id check in the guard is redundant, but wasn't sure how else to accomplish this. 😕

RavenHursT: Duh.. guess I should've made a test for this eh? 🤦 

https://goodpartyorg.postman.co/workspace/66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/34024107-a40b0524-eaaf-4491-b83b-68789501be27

oldfashionedrobot (approved):

PR #39, opened 2025-01-03
[WEB-3388] - checkout session

RavenHursT:
https://goodparty.atlassian.net/browse/WEB-3387

[Test Run](https://goodpartyorg.postman.co/workspace/66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/34024107-0025b2e5-aeb3-4e84-9706-d4614499ac0e)


src/types.ts
```diff
@@ -0,0 +1,12 @@
+export {}
+
+// TODO: figure out if it's possible to collocate these Json types with the
+//  models they're associated with
+declare global {
+  export namespace PrismaJson {
```
oldfashionedrobot, line 6: You should be able to put this file in the `/prisma` folder, I think our tsconfig by default will pick up any .ts files in the project directory.  

I _think_ with [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html), we may be able to split these JSON typedefs into multiple files, maybe in a `prisma/types` folder or similar

oldfashionedrobot (commented):
Looks good to me, had one comment about the types.ts file

src/types.ts
RavenHursT: Looks like this works well.. 

Gonna move the file.. I'll attempt to also split the types out. If that doesn't work quickly, I'll add a `TODO` to get that figured out later on

RavenHursT: Worked!

https://github.com/thegoodparty/gp-api/pull/39/commits/59dd06a917120d5634fa4901c14127daeef88c9f

oldfashionedrobot (approved):

PR #41, opened 2025-01-06
WEB-3362: Nestjs - migrate campaign/launch

oldfashionedrobot:
Migrates campaign/launch endpoint

Postman tests: https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/38094541-b44771d1-d399-4b87-b784-42f2fd75d515

### DIFF
--- a/src/campaigns/campaigns.controller.ts
+++ b/src/campaigns/campaigns.controller.ts
@@ -99,12 +99,6 @@ export class CampaignsController {
       this.logger.error('Error at campaign launch', e)
       await this.slack.errorMessage('Error at campaign launch', e)
 
-      if (e instanceof Error) {
-        throw new BadRequestException(
-          'Cannot launch campaign without an Office',
-        )
-      }
-
       throw e
     }
   }
--- a/src/campaigns/services/campaigns.service.ts
+++ b/src/campaigns/services/campaigns.service.ts
@@ -1,4 +1,4 @@
-import { Injectable, Logger } from '@nestjs/common'
+import { BadRequestException, Injectable, Logger } from '@nestjs/common'
 import { PrismaService } from '../../prisma/prisma.service'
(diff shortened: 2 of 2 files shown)


src/campaigns/campaigns.controller.ts
```diff
@@ -78,4 +88,24 @@ export class CampaignsController {
...
+
+      if (e instanceof Error) {
+        throw new BadRequestException(
+          'Cannot launch campaign without an Office',
+        )
+      }
+
+      throw e
+    }
+  }
```
RavenHursT, line 110: I don't think this is what we want here.. All exceptions from the service are going to be instances of prototypes of `Error`:
image

Prolly better to just throw the `BadRequestException` down in the service, here:  https://github.com/thegoodparty/gp-api/pull/41/files#diff-ffce57cc5ae4181953538cc126a97baa019723136df9a25e9d94154b56b639e8R215  ...and delete this if statement all together.

RavenHursT: Oh.. you'll def still have to `throw` the exception again so that Nest can handle it accordingly tho..

src/campaigns/services/campaigns.service.ts
```diff
@@ -297,6 +346,23 @@ export class CampaignsService {
...
+          link: `${APP_BASE}/dashboard`,
+        },
+      })
+    } catch (e) {
+      this.logger.error('Error sending campaign launch email', e)
+    }
+  }
 }
 
 // async function claimExistingCampaignRequests(user, campaign) {
```
RavenHursT, line 368: we can delete this since we're not going to be migrating the Campaign Manager stuff

RavenHursT (changes requested):

RavenHursT (approved):

PR #45, opened 2025-01-08
WEB-3423: Nestjs: Migrate admin victory mail

oldfashionedrobot:
- Migrates the `admin/victory-mail` endpoint to `admin/campaigns/:id/send-victory-email`  - [Webapp migration subtask here](https://goodparty.atlassian.net/browse/WEB-3483)
- Added `pathToVictory.jsonTypes.ts` with some fields for typing `pathToVictory.data`
- Cleaned up duplicated functionality in `adminCampaigns.service.ts`


src/admin/campaigns/adminCampaigns.service.ts
```diff
@@ -101,50 +80,45 @@ export class AdminCampaignsService {
...
+      { user: true, pathToVictory: true },
+    )
+    const { pathToVictory, user } = campaign
+
+    if (!pathToVictory) {
+      throw new Error('Path to Victory is not set.')
+    }
+    if (!user) {
+      throw new Error('Campaign has no user')
+    }
```
RavenHursT, line 103: Let's make both of these `throw new BadRequestException()`

src/campaigns/services/campaigns.service.ts
```diff
@@ -235,6 +244,26 @@ export class CampaignsService {
     return true
   }
 
+  async findSlug(user: User, suffix?: string) {
```
RavenHursT, line 247: Let's call this `findBySlug`.. this reads like we're finding a slug, when in fact, we're finding a campaign, by it's slug

RavenHursT (changes requested):

src/campaigns/services/campaigns.service.ts
oldfashionedrobot: This one actually is finding a slug, I had originally made it a utility function that took a prisma client as an arg, but moved it here, since it is a campaign concern

RavenHursT: Oh.. duh.. ok.. total 🧠 💨  on my behalf.. 👍

RavenHursT (approved):

PR #47, opened 2025-01-10
Web 3361

oakinh:
- Adds a mapping service and controller under module
- Adds additional fields to campaign details and data
- Major changes to campaign factory and seeding, including adding pre-made campaigns (this is done to ensure at least some campaigns fit a complex criteria for the mapping endpoints, which should never return 0 in the Postman tests I made)
- Moves race factories and seeding into the proper folder
- Adds a couple shared util functions

https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/38601784-f7738059-b732-45aa-b130-352796b88dbf


src/campaigns/mapping/mapping.controller.ts
```diff
@@ -0,0 +1,26 @@
+import { Controller, Get, Query } from '@nestjs/common'
+import { MappingService } from './mapping.service'
+import { CleanCampaign } from '../campaigns.types'
+import { MapCountDto, MapDto } from '../schemas/mappingSchemas'
+
+@Controller('campaigns/mapping')
+export class MappingController {
```
oldfashionedrobot, line 7: This is a little nitpicky, but I don't know if `Mapping` is the best name here, these are providing campaign data to show on a map, not generating maps or mapping data to other data. I think something like `CampaignMapController` / `CampaignsMapService` is more descriptive of what it is for.

src/campaigns/mapping/mapping.service.ts
```diff
@@ -0,0 +1,221 @@
...
+            details: {
+              path: ['electionDate'],
+              gte: sevenDaysAgo,
+            },
+          },
+        ],
+      },
+    ]
+
+    const combinedAndConditions: Prisma.CampaignWhereInput[] = [
```
oldfashionedrobot, line 54: You can just `.concat` these arrays together, spreading makes unnecessary clones of them.

src/campaigns/util/buildMapFilters.ts
```diff
@@ -0,0 +1,144 @@
...
+
+interface FilterParams {
+  partyFilter?: string
+  stateFilter?: string
+  levelFilter?: string
+  resultsFilter?: boolean
+  officeFilter?: string
+  nameFilter?: string
+  forceReCalc?: boolean
+  isProd?: boolean
```
oldfashionedrobot, line 12: `isProd` doesn't need to be an arg here, since it pulls from an env variable, you can just pull that env var and check it here.

src/shared/util/dates.util.ts
```diff
@@ -0,0 +1,3 @@
+export const sevenDaysAgo = new Date(Date.now() - 7 * 24 * 60 * 60 * 1000)
```
oldfashionedrobot, line 1: We have the `date-fns` package in the project, you can use that to do something like `subDays(new Date(), 7)` to get a date 7 days ago

tsconfig.json
```diff
@@ -18,6 +18,7 @@
     "strictBindCallApply": false,
     "forceConsistentCasingInFileNames": false,
     "noFallthroughCasesInSwitch": false,
-    "lib": ["esnext"]
+    "lib": ["esnext"],
+    "resolveJsonModule": true,
```
oldfashionedrobot, line 22: Shouldn't need this if you make the change in my top comment

src/campaigns/util/geoLocation.ts
```diff
@@ -0,0 +1,69 @@
...
+  details: PrismaJson.CampaignDetails,
+  prismaService: PrismaService,
+): Promise<{ lat: number; lng: number; geoHash: string } | null> {
+  if (!details?.zip || !details?.state) return null
+
+  const globalCoords = await zipToLatLng(details?.zip, details?.state)
+  if (globalCoords == null) return null
+
+  const { lat, lng, geoHash } = globalCoords
+  await prismaService.campaign.update({
```
oldfashionedrobot, line 52: These geocoding function along with zipToLatLng should all probably be in a `GeocodingService`. Here, you'd need to inject the `CampaignsService` and use it to update a campaign instead of taking in the PrismaService as an argument

oldfashionedrobot (changes requested):

prisma/schema/campaign.jsonTypes.ts
```diff
@@ -22,24 +22,52 @@ declare global {
...
+      slug?: string | null
       currentStep?: OnboardingStep
-      slug?: string
+      launchStatus?: CampaignLaunchStatus
+      customIssues?: Record<'title' | 'position', string>[]
+      runningAgainst?: Record<'name' | 'party' | 'description', string>[]
+      party?: string //TODO: enumerate all parties we want to allow?
+      otherParty?: string
+      office?: string
+      otherOffice?: string
```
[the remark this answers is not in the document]
RavenHursT: Yeah.. something's def not right w/ these fields:

image

src/campaigns/mapping/mapping.service.ts
RavenHursT: Or just create them in place instead of storing them in a variable that you're not reading from anywhere else anyway... 🤷 

```suggestion
const combinedAndConditions: Prisma.CampaignWhereInput[] = [
      ...buildMapFilters({
        partyFilter,
        stateFilter,
        levelFilter,
        resultsFilter,
        officeFilter,
        isProd,
      }),
      {
        OR: [
          { didWin: true },
          {
            didWin: null,
            details: {
              path: ['electionDate'],
              gte: sevenDaysAgo,
            },
          },
        ],
      },
    ]
```

src/campaigns/util/buildMapFilters.ts
RavenHursT: See my comment above regarding `isPro` and the `appEnvironment` in general..

prisma/schema/campaign.jsonTypes.ts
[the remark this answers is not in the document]
oakinh: I'm not following, I don't see the issue you're seeing in your editor, and I don't think I added most of those to campaignData, a lot of that came in when I was trying to fix the  conflicts when merging with dev. And normalizedOffice is in campaign details for me

seed/campaigns.ts
```diff
@@ -5,10 +5,15 @@ import { userFactory } from './factories/user.factory'
 import { pathToVictoryFactory } from './factories/pathToVictory.factory'
 import { buildSlug } from 'src/shared/util/slug.util'
 import { getFullName } from 'src/users/util/users.util'
+import { Campaign } from '@prisma/client'
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const fixedCampaigns = require('./fixedCampaigns.json')
```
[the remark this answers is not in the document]
oakinh: I couldn't get this working because I was putting it in curly braces, but this works, thanks

src/campaigns/util/zipToLatLng.ts
```diff
@@ -0,0 +1,39 @@
+import * as ngeohash from 'ngeohash'
+
+const googleApiKey = process.env.GOOGLE_API_KEY
```
RavenHursT, line 3: This env var needs to be added to the `.env.example`

prisma/schema/campaign.jsonTypes.ts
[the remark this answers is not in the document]
oldfashionedrobot: It may have been because of a merge conflict. If you look at: https://github.com/thegoodparty/gp-api/blob/develop/prisma/schema/campaign.jsonTypes.ts, the file should probably just be what's there + what fields you needed to add in this PR

src/campaigns/mapping/mapping.service.ts
```diff
@@ -0,0 +1,221 @@
...
+      include: {
+        user: {
+          select: {
+            firstName: true,
+            lastName: true,
+            avatar: true,
+          },
+        },
+      },
+    })
```
RavenHursT, line 139: This should be going through the `CampaignsService`, not directly to the ORM.

```suggestion
    const campaigns = await this.campaignsService.findAll({
      where,
      include: {
        user: {
          select: {
            firstName: true,
            lastName: true,
            avatar: true,
          },
        },
      },
    })
```

RavenHursT (changes requested):
Found something else...

prisma/schema/campaign.jsonTypes.ts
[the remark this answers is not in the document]
RavenHursT: Weird.. must've been my IDE... those errors went away now...

I'm still pretty sure that those duplicate fields don't belong on `CampaignData`..

src/campaigns/mapping/mapping.service.ts
oakinh: I guess I should change that for all my campaign prisma use then, because there's more than just that spot. I am a little confused though why our campaign service methods take different arguments than the original prisma method, doesn't work the best for what I'm trying to do in some spots, because I'm using Prisma's actual types

tsconfig.json
oakinh: I changed the import to Matthew's suggestion so I don't have to use the ES-lint ignore. I like keeping the file as a json just for the simple reason it's easy to visually distinguish with the icon from the rest of the directory's contents

src/campaigns/mapping/mapping.service.ts
oakinh: Seems a little clunky to me spreading the result of a function inside an array creation, but alright I made the change

src/campaigns/campaigns.types.ts
```diff
@@ -11,12 +11,32 @@ export enum CampaignLaunchStatus {
...
+
 export enum CampaignStatus {
   candidate = 'candidate',
   onboarding = 'onboarding',
 }
 
-export enum OnboardingStep {
-  complete = 'onboarding-complete',
-  registration = 'registration',
+export interface CleanCampaign {
```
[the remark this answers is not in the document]
oakinh: Basically all of this can't be picked since it's in JSON but I made the change

seed/campaigns.ts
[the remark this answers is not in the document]
oakinh: Actually this doesn't work, it doesn't recognize it as an array, only if I use the require style import

oakinh:
I've had to do merge with dev and resolve conflicts twice now, so I'm going to wait until this is semi-approved before doing it a third time

oldfashionedrobot (changes requested):

prisma/schema/campaign.jsonTypes.ts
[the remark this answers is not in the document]
RavenHursT: I'm still seeing a lot of properties on `CampaignData` that AFAICT are indeed supposed to be on `CampaignDetails` 🤔 

image

seed/campaigns.ts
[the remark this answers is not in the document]
RavenHursT: This should be good to go w/ that tsconfig change

seed/campaigns.ts
```diff
@@ -57,3 +61,34 @@ export default async function seedCampaigns(
...
+    user = await prisma.user.create({
+      data: {
+        ...userData,
+        metaData:
+          userData.metaData !== null ? userData.metaData : Prisma.JsonNull,
+      },
+    })
+  }
+  return user
+}
```
RavenHursT, line 81: Yeah.. services w/ responsibility over one resource, should _not_ be going directly to the ORM to perform transactions w/ other types of resources.

This needs to be a call to the creation method in the `UsersService`, not directly to Prisma.

seed/campaigns.ts
```diff
@@ -57,3 +61,34 @@ export default async function seedCampaigns(
...
+    const userData = userFactory()
+    user = await prisma.user.create({
+      data: {
+        ...userData,
+        metaData:
+          userData.metaData !== null ? userData.metaData : Prisma.JsonNull,
+      },
+    })
+  }
+  return user
```
RavenHursT, line 80: Why the redundant variable?

seed/campaigns.ts
```diff
@@ -57,3 +61,34 @@ export default async function seedCampaigns(
 
   return campaignIds
 }
```
RavenHursT, line 63: Let's see if we can't get this down to a single for-loop, or `.map()` expression calling a _single_ method.

prisma/schema/campaign.jsonTypes.ts
```diff
@@ -22,24 +22,48 @@ declare global {
...
+      launchStatus?: CampaignLaunchStatus
+      customIssues?: Record<'title' | 'position', string>[]
+      runningAgainst?: Record<'name' | 'party' | 'description', string>[]
+      website?: string
+      district?: string
+      pastExperience?: string
+      occupation?: string
+      funFact?: string
+      campaignCommittee?: string
+      statementName?: string
```
RavenHursT, line 66: There's still a lot of fields on here that are not on `.data` but on `.details` in the current app.

src/campaigns/util/buildMapFilters.ts
```diff
@@ -0,0 +1,138 @@
...
+    details: {
+      path: ['zip'],
+      not: { equals: null },
+    },
+    didWin: {
+      not: false,
+    },
+  })
+
+  if (isProd) {
```
RavenHursT, line 92: Add `TODO`s ont he `isProd`s here w/ a mention of the ticket here: https://goodparty.atlassian.net/browse/WEB-3487

src/campaigns/mapping/mapping.service.ts
RavenHursT: Yeah.. def go ahead and refactor those to allow for us to have the same arguments interface to the Prisma model methods, if it makes sense to do so.

RavenHursT (changes requested):
Just a few more things.

seed/campaigns.ts
oldfashionedrobot: This is just a seed file, I don't know if we need to make Nikao clean it up for this PR

RavenHursT: Oh.. I spoke w/ @oakinh on Slack and presented it to him as a "if you want a challenge..." kind of thing...

seed/campaigns.ts
```diff
@@ -5,9 +5,14 @@ import { userFactory } from './factories/user.factory'
 import { pathToVictoryFactory } from './factories/pathToVictory.factory'
 import { buildSlug } from 'src/shared/util/slug.util'
 import { getFullName } from 'src/users/util/users.util'
-
-const NUM_CAMPAIGNS = 20
+import { Campaign } from '@prisma/client'
+// eslint-disable-next-line
+//const fixedCampaigns = require('./fixedCampaigns.json') // Normal import doesn't correctly recognize it as an array
+import fixedCampaigns from './fixedCampaigns.json'
```
oldfashionedrobot, line 11: Instead of doing these workaround and disabling eslinting, you can just rename `fixedCampaigns.json` to `fixedCampaigns.ts`, and inside that file export your json array as a variable, like `export const FIXED_CAMPAIGNS = [{......data}]`

Then you can just import that variable and not have to disable eslint and retype your import.

seed/fixedCampaigns.json
```diff
@@ -0,0 +1,233 @@
+[
```
oldfashionedrobot, line 1:
```suggestion
export const FIXED_CAMPAIGNS = [
```

oldfashionedrobot: (and will also have to change the file type to `.ts` instead of `.json`

seed/campaigns.ts
oldfashionedrobot:
```suggestion
import { FIXED_CAMPAIGNS } from './fixedCampaigns.ts'
```
(after making the change below: https://github.com/thegoodparty/gp-api/pull/47/files#r1916710263)

seed/campaigns.ts
RavenHursT: Please ignore me.

src/campaigns/campaigns.module.ts
```diff
@@ -2,19 +2,34 @@ import { forwardRef, Module } from '@nestjs/common'
...
+import { RacesModule } from 'src/races/races.module'
+import { RacesService } from 'src/races/races.service'
+import { GraphqlModule } from 'src/graphql/graphql.module'
+import { GraphqlService } from 'src/graphql/graphql.service'
 
 @Module({
-  imports: [EmailModule, UsersModule, forwardRef(() => CampaignsAiModule)],
-  controllers: [CampaignsController, CampaignPositionsController],
+  imports: [EmailModule, UsersModule, RacesModule, forwardRef(() => CampaignsAiModule)],
+  //imports: [EmailModule, UsersModule, forwardRef(() => CampaignsAiModule)],
```
oldfashionedrobot, line 20: Found a leftover comment

src/campaigns/map/campaignMap.service.ts
```diff
@@ -0,0 +1,211 @@
...
+            OR: [
+              { firstName: { contains: nameFilter, mode: 'insensitive' } },
+              { lastName: { contains: nameFilter, mode: 'insensitive' } },
+            ],
+          },
+        ],
+      }
+    }
+
+    const campaigns = await this.prisma.campaign.findMany({
```
oldfashionedrobot, line 119: Probably can just be `this.campaignsService.findAll`?

src/races/races.module.ts
```diff
@@ -7,5 +7,6 @@ import { GraphqlModule } from 'src/graphql/graphql.module'
   controllers: [RacesController],
   providers: [RacesService],
   imports: [GraphqlModule],
+  exports: [RacesModule]
```
oldfashionedrobot, line 10: I don't think you need to export the whole module here, just the `RacesService`

src/campaigns/campaigns.module.ts
```diff
@@ -2,19 +2,34 @@ import { forwardRef, Module } from '@nestjs/common'
...
+import { CampaignMapController } from './map/campaignMap.controller'
+import { CampaignMapService } from './map/campaignMap.service'
 import { CampaignPlanVersionsService } from './services/campaignPlanVersions.service'
 import { EmailModule } from 'src/email/email.module'
 import { CampaignPositionsController } from './positions/campaignPositions.controller'
 import { CampaignPositionsService } from './positions/campaignPositions.service'
 import { UsersModule } from 'src/users/users.module'
+import { GeocodingService } from './services/geocoding.service'
+import { RacesModule } from 'src/races/races.module'
+import { RacesService } from 'src/races/races.service'
```
oldfashionedrobot, line 14: If you change the export in `RacesModule` to export the `RacesService` you just need to import the `RacesModule` here, you shouldn't have to add the `RacesService` as a provider here. 

Also, doesn't look like the Grapql imports here are used.

src/races/races.service.ts
```diff
@@ -25,7 +25,7 @@ export class RacesService {
   private readonly logger = new Logger(RacesService.name)
   constructor(
     private prisma: PrismaService,
-    private graphQLService: GraphqlService,
+    @Optional() private graphQLService: GraphqlService,
```
oldfashionedrobot, line 28: Does this`@Optional()` need to be here? You may have had to add it here to be able to register this service as a provider  in the `CampaignsModule`. If you make the change to the `RacesModule` exports, I don't think you'll need this decorator

oldfashionedrobot (changes requested):
Found some much smaller things, but overall is looking great.

seed/campaigns.ts
```diff
@@ -1,59 +1,143 @@
...
+      const { campaignId, fakeP2V, fakeUpdateHistory } = creationData
+
+      campaignIds.push(campaignId)
+      fakeP2Vs.push(fakeP2V)
+      fakeUpdateHistory.push(...fakeUpdateHistory)
     }
   }
 
-  await prisma.pathToVictory.createMany({ data: fakeP2Vs })
+  //await prisma.pathToVictory.createMany({ data: fakeP2Vs })
```
RavenHursT, line 71: delete

seed/campaigns.ts
```diff
@@ -1,59 +1,143 @@
...
-          userId: user.id,
-        }),
+  for (let i = 0; i < loopLength; i++) {
+    if (i < FIXED_CAMPAIGNS.length) {
+      const creationData = await createCampaignAndUser(
+        existingUsers,
+        prisma,
+        FIXED_CAMPAIGNS[i],
       )
+      const { campaignId, fakeP2V, fakeUpdateHistory } = creationData
```
RavenHursT, line 55: You can do this destructuring on the same line as the function call above. No need for the extra variable.

seed/campaigns.ts
```diff
@@ -1,59 +1,143 @@
...
+}
 
 export default async function seedCampaigns(
   prisma: PrismaClient,
   existingUsers: User[],
 ) {
-  const fakeP2Vs: any[] = []
-  const fakeUpdateHistory: any[] = []
-
+  const fakeP2Vs: FakeP2V[] = []
```
RavenHursT, line 42: I don't see you using this anywhere other than filling it up?

image

Seems like you could just get rid of it here as well as from the return from `createCampaignAndUser` 🤔

seed/campaigns.ts
```diff
@@ -1,59 +1,143 @@
...
+      }),
+    })
+  } else {
+    campaign = await prisma.campaign.create({
+      data: campaignFactory({
+        userId: user.id,
+        slug: buildSlug(getFullName(user)),
+      }),
+    })
+  }
```
RavenHursT, line 105: You could just use a ternary here:

```suggestion
  const campaign =  await prisma.campaign.create({
      data: campaignFactory({
        userId: user.id,
        slug: buildSlug(getFullName(user)),
        ...(fixedData || {}),
      }),
    })
```

seed/campaigns.ts
RavenHursT: This one not addressed?

RavenHursT (changes requested):
Just a couple minor things.. hopefully this is it 🤞 

Great job powerin' through this man! 💪

oakinh:
Since a lot of changes have been made, here's another test run just in case

https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/38601784-eae1e01d-21ab-4f86-b75b-770133062aea

src/campaigns/campaigns.module.ts
```diff
@@ -2,19 +2,30 @@ import { forwardRef, Module } from '@nestjs/common'
...
+    CampaignPositionsController,
+    CampaignMapController,
+  ],
   providers: [
     CampaignsService,
     CampaignPlanVersionsService,
     CampaignPositionsService,
+    CampaignMapService,
+    GeocodingService,
+    RacesService,
```
oldfashionedrobot, line 28: Instead of putting the `RacesService` here as a provider, you want to add the `RacesModule` above as an imported module. 

Now that you have the `RacesService` exported from the `RacesModule`, when you import the `RacesModule` here, you'll have access to the `RacesService` inside the `CampaignMapService` and any other campaign module classes.

After that, I think you should be able to remove the `@Optional` decorator from the `RacesService` constructor

oldfashionedrobot (approved):

RavenHursT (approved):

PR #51, opened 2025-01-14
feat: voters service.

taylorgp:
- Adds a VoterService for querying the L2 api

Sanjeev Adds
- Combined voters.module + votersService with voterData.module controllers/services, so a lot of changes are updated import paths
- Added remaining types to voterService

Postman VoterData tests: https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/38094541-c218caa4-0141-4f21-a7cd-b9dc0151b1fa


src/voters/voters.service.ts
```diff
@@ -0,0 +1,566 @@
...
+    }
+    if (columnsResponse?.data?.columns) {
+      columns = columnsResponse.data.columns
+    }
+    return columns
+  }
+
+  async getPartisanCounts(
+    electionState: string,
+    searchJson: any,
```
oldfashionedrobot, line 277: It looks like this is at least expected to be an object, `object` might be better than just `any`

src/voters/voters.service.ts
```diff
@@ -0,0 +1,566 @@
...
+  async getEstimatedCounts(
+    electionState: string,
+    searchJson: any,
+  ): Promise<number> {
+    let count = 0
+    // Note: this endpoint also returns # of households which we don't use.
+    // This endpoint could use same query as getPartisanCounts but we use a different endpoint
+    // but if we need partisan election counts we can use the same endpoint.
+    const searchUrl = `${API_BASE}/records/search/estimate/1OSR/VM_${electionState}?id=1OSR&apikey=${L2_DATA_KEY}`
+    let response: any
```
oldfashionedrobot, line 326: the response var could probably be `unknown`, or at least you can leave off the explicit `any` typing, since it should implicitly be any already

src/voters/voters.service.ts
```diff
@@ -0,0 +1,566 @@
...
+      )
+    }
+    return turnoutDates
+  }
+
+  private determineHistoryColumn(
+    electionDate: string,
+    electionState: string,
+    yearOffset: number,
+    columns: any[],
```
oldfashionedrobot, line 482: Could columns get a type here?

RavenHursT (approved):

PR #55, opened 2025-01-17
[WEB-3440] - Generate File Upload Url

RavenHursT:
https://goodparty.atlassian.net/browse/WEB-3440

Postman run: https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/34024107-c3f07836-75d5-4d93-9e6f-cd2077c8ed9a


src/users/users.types.ts
```diff
@@ -0,0 +1,5 @@
+export type GenerateSignedUploadUrlArgs = {
```
oldfashionedrobot, line 1: Do you need this type? You should be able to use that `GenerateSignedUploadUrlArgsDto` as this same exact type. Or alias it as  `GenerateSignedUploadUrlArgs` if you need the different name

.env.example
```diff
@@ -11,6 +11,8 @@ SQS_QUEUE_URL=https://sqs.us-west-2.amazonaws.com/ACCOUNT_ID/QUEUE_NAME.fifo
 AWS_REGION=us-west-2
 AWS_ACCESS_KEY_ID=super-secret-id
 AWS_SECRET_ACCESS_KEY=super-secret-key
+AWS_S3_KEY=super-secret-s3-key
+AWS_S3_SECRET=super-secret-s3-secret
```
oldfashionedrobot, line 15: Are these `AWS_S3` keys used?

oldfashionedrobot (commented):
Just found two small things

.env.example
RavenHursT: They will be...
image

src/users/users.types.ts
RavenHursT: 🤔

RavenHursT: But the DTO doesn't give me the same level of introspection:
image

vs

image

I really wish I could reuse the DTO, but I also want to be able to see what I'm workin' with, w/o having to go open that other file 🤷

oldfashionedrobot (approved):

PR #58, opened 2025-01-22
WEB-3428: Nestjs: voter file - schedule

oldfashionedrobot:
- Migrates voter file campaign scheduling 
- Also adds a file mimeType validation to the `FilesInterceptor`

### DIFF
--- a/src/files/interceptors/files.interceptor.ts
+++ b/src/files/interceptors/files.interceptor.ts
@@ -3,6 +3,7 @@ import {
   ExecutionContext,
   CallHandler,
   Logger,
+  BadRequestException,
 } from '@nestjs/common'
 import { FastifyRequest } from 'fastify'
 import { FileUpload } from '../files.types'
@@ -24,6 +25,8 @@ type FilesInterceptorOpts = {
   numFiles?: number
   /** Maximum file size to allow, an error will be thrown if files are larger. */
   sizeLimit?: number
+  /** Array of mimetypes to accept, an error will be thrown if any files do not match. */
+  mimeTypes?: string[]
 }
 
 /**
@@ -44,7 +47,12 @@ type FilesInterceptorOpts = {
  */
 export function FilesInterceptor(
   key: string = 'file',
-  { mode = 'buffer', numFiles, sizeLimit }: FilesInterceptorOpts = {},
+  {
+    mode = 'buffer',
+    numFiles,
+    sizeLimit,
+    mimeTypes,
+  }: FilesInterceptorOpts = {},
 ) {
   return class MixinInterceptor implements NestInterceptor {
     logger = new Logger(FilesInterceptor.name)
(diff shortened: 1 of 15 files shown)


src/files/interceptors/files.interceptor.ts
```diff
@@ -85,6 +102,7 @@ export function FilesInterceptor(
             ...omit(part, ['file', 'toBuffer', 'type', 'fields']),
           })
         } else {
+          // set multipart fields directly on request body
           req.body[part.fieldname] = part.value
         }
```
RavenHursT, line 107: Do we need to do the checks above for the multipart use case as well? 🤔

src/voterData/util/voterOutreach.util.ts
```diff
@@ -0,0 +1,444 @@
...
+              ],
+            },
+          ],
+        }
+      : undefined,
+    imageUrl
+      ? {
+          type: 'section',
+          text: {
+            type: 'mrkdwn',
```
RavenHursT, line 433: Please change the `type` string constants here to reference the `SlackMessageType` enum here: https://github.com/thegoodparty/gp-api/blob/543f5197e810b3a8c1a37b5c7cd7a5ebdc802ab5/src/shared/services/slackService.types.ts#L11-L22

src/voterData/voterOutreach.service.ts
```diff
@@ -0,0 +1,141 @@
...
+    // if (!crmCompany) {
+    //   throw new Error(`crmCompany not found for ${campaign}`)
+    // }
+    // const assignedPa = await getCrmCompanyOwnerName(crmCompany, true)
+
+    const messagingScript: string = campaign.aiContent?.[script]?.content
+      ? sanitizeHtml(campaign.aiContent?.[script]?.content, {
+          allowedTags: [],
+          allowedAttributes: {},
+        })
```
RavenHursT, line 51: Shouldn't this be getting sanitized on the way _into_ the DB, not on the way out? 🤔

src/voterData/voterOutreach.service.ts
```diff
@@ -0,0 +1,141 @@
...
+          elements: [
+            {
+              type: 'text',
+              text: `${key}: `,
+              style: {
+                bold: true,
+              },
+            },
+            {
+              type: 'text',
```
RavenHursT, line 88: `SlackMessageType`s here as well

src/voterData/voterOutreach.service.ts
```diff
@@ -0,0 +1,141 @@
...
+      imageUrl,
+      message,
+      formattedAudience,
+      audienceRequest: audience['audience_request'],
+    })
+
+    await this.slack.message(
+      slackBlocks,
+      IS_PROD ? SlackChannel.botPolitics : SlackChannel.botDev,
+    )
```
RavenHursT, line 123: Don't _think_ we need to `await` this here.. we can just fire and forget most likely..

src/files/interceptors/files.interceptor.ts
oldfashionedrobot: No, the above part is for files found in the mulitpart data, this is for fields. The fields just get added onto the body of the request, and our regular zod schema stuff takes over from there

src/voterData/voterOutreach.service.ts
oldfashionedrobot: This is just sanitizing it before its sent over to slack, it might not even be necessary, but looks like this sanitizeHtml has been in the tgp-api version for some time

src/voterData/voterOutreach.service.ts
oldfashionedrobot: I think we actually do want it here, because if this errors we don't really need to update the campaign below. Right now the only record of these schedule requests is the slack message, so if that fails the whole thing should fail.

src/voterData/voterOutreach.service.ts
oldfashionedrobot: I threw this slack block building stuff in quickly, but there are probably packages out there to help compose slack block messages, so maybe we can replace a lot of this in the near future

RavenHursT (approved):

PR #59, opened 2025-01-22
Web 3433

oakinh:
This just migrates this endpoint - https://github.com/thegoodparty/tgp-api/blob/develop/api/controllers/content/blog-articles-by-section.js

This takes advantage of the pre-existing findByType method, and doesn't create another service method.

https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/38601784-677b068c-a563-4d5c-b40f-2fbff2126334

The titles test that failing is just because I didn't merge that recent PR into my branch yet, and the fetch all test that's failing it's literally off by 1 for some reason unrelated to what I'm doing I believe

### DIFF
--- a/src/content/content.controller.ts
+++ b/src/content/content.controller.ts
@@ -1,4 +1,10 @@
-import { BadRequestException, Controller, Get, Param } from '@nestjs/common'
+import {
+  BadRequestException,
+  Controller,
+  Get,
+  Param,
+  InternalServerErrorException,
+} from '@nestjs/common'
 import { ContentService } from './content.service'
 import { ContentType } from '@prisma/client'
 import {
@@ -10,6 +16,7 @@ import {
   mapGlossaryItemsToSlug,
 } from './util/glossaryItems.util'
 import { PublicAccess } from '../authentication/decorators/PublicAccess.decorator'
+import { BlogArticleAugmented, BlogSection, Hero } from './content.types'
 
 @Controller('content')
 @PublicAccess()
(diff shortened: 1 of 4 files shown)


src/content/content.controller.ts
```diff
@@ -67,4 +74,78 @@ export class ContentController {
...
+          hero = { id, title, mainImage, publishDate, slug, summary }
+          section.articles = section.articles.slice(1)
+          results.push(section)
+        } else {
+          delete section.articles
+          results.push(section)
+        }
+      }
+      return { sections: results, hero: hero, sectionIndex }
+    } else {
```
oldfashionedrobot, line 114: Instead of doing this if/else here (that I know was already in the tgp-api code), You could split this into two separate endpoints. One for `blog-articles-by-section/:sectionSlug` and one for just `blog-articles-by-section`. 

That way, if nest sees `blog-articles-by-section` with a section slug appended, it will route to the right handler instead of you having to do it manually

oakinh: Great idea, I knew there was a better pattern here. Will implement, thank you!

src/content/content.controller.ts
```diff
@@ -67,4 +74,77 @@ export class ContentController {
...
     }
   }
+
+  @Get('blog-articles-by-section/:sectionSlug')
+  async findBlogArticlesBySection(@Param('sectionSlug') sectionSlug: string) {
+    const sections: BlogSection[] = await this.contentService.findByType(
+      InferredContentTypes.blogSections,
+    )
+    if (!sections) {
+      throw new InternalServerErrorException("Blog sections couldn't be pulled")
```
RavenHursT, line 84: Did this pass the linter?  We should be enforcing single quotes last I checked... 🤔

RavenHursT: Oh.. maybe it's the `couldn't` in there...

RavenHursT (changes requested):
I'm seeing failures in your linked Postman run? 🤔 

image

src/content/content.controller.ts
oakinh: No linter warnings on this one, I can use ticks if you prefer for the couldn't

RavenHursT: Meh.. if it's passing the linter, I really don't care. Was just surprised to see the `"`s is all.

RavenHursT (approved):

PR #61, opened 2025-01-22
[WEB-3431] - Full Story Integration

RavenHursT:
https://goodparty.atlassian.net/browse/WEB-3431

Postman run: https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/34024107-ca220342-717e-4da1-a9d0-85af52df6769

Results log:
image


src/app.module.ts
```diff
@@ -44,7 +46,9 @@ import { FilesModule } from './files/files.module'
     PaymentsModule,
     ErrorLoggerModule,
     VoterDataModule,
+    FullStoryModule,
     FilesModule,
+    AwsModule,
```
oldfashionedrobot, line 51: I don't think these two modules need to be imported here.

src/campaigns/services/campaigns.service.ts
```diff
@@ -107,6 +107,11 @@ export class CampaignsService {
   }
 
   async update(args: Prisma.CampaignUpdateArgs) {
+    const user = await this.usersService.findUser(
+      args.where as Prisma.UserWhereUniqueInput,
```
oldfashionedrobot, line 111: Is this correct? You're using the same where arguments for finding a user and in the campaign update? 

You could also just query the campaign and pass the `userId` instead of getting the user here too I think

src/campaigns/services/campaigns.service.ts
```diff
@@ -155,11 +160,9 @@ export class CampaignsService {
...
       // }
 
-      // try {
-      //   await sails.helpers.fullstory.customAttr(user.id)
-      // } catch (e) {
-      //   sails.helpers.log(campaign.slug, 'error updating fullstory', e)
-      // }
+      const user = await this.usersService.findUser({ id: campaign.userId })
+
+      user && (await this.fullstoryService.trackByUserId(user.id))
```
oldfashionedrobot, line 165: These findUser and trackByUserId calls could probably be bundled into a method in the UsersService that just takes in a userId. The Fullstory service maybe doesn't need to be in here.

oldfashionedrobot: Also the fullstory service has the user service injected, you could just move the findUser part into the trackByUserId method

src/fullStory/fullStory.controller.ts
```diff
@@ -0,0 +1,25 @@
+import { Controller, Get, HttpCode, HttpStatus } from '@nestjs/common'
+import { FullStoryService } from './fullStory.service'
+import { ReqUser } from '../authentication/decorators/ReqUser.decorator'
+import { User } from '@prisma/client'
+import { CampaignWith } from '../campaigns/campaigns.types'
+import { CampaignsService } from '../campaigns/services/campaigns.service'
+
+@Controller('integrations')
```
oldfashionedrobot, line 8: was this intended to still be `integrations`?

src/fullStory/fullStory.controller.ts
```diff
@@ -0,0 +1,25 @@
...
+    private readonly fullstory: FullStoryService,
+    private readonly campaigns: CampaignsService,
+  ) {}
+
+  @Get('fullstory-sync')
+  @HttpCode(HttpStatus.ACCEPTED)
+  async syncFullStoryUsers() {
+    // No need for await here to return immediately
+    this.fullstory.trackCampaigns(
+      (await this.campaigns.findAll({
```
oldfashionedrobot, line 20: I think if you're going to inject the campaigns service and user service into the fullstory service, It'd be cleaner to do any querying for stuff inside the fullstory service right? Instead of querying then passing into this method?

src/fullStory/fullStory.service.ts
```diff
@@ -0,0 +1,291 @@
...
+        failed: 0,
+      },
+      this.multiTrackIterator.bind(this),
+    )
+    this.logger.log('FullStory trackCampaigns results:', resultCounts)
+    return resultCounts
+  }
+
+  async trackByUserId(userId: number) {
+    const user = await this.users.findUser({ id: userId })
```
oldfashionedrobot, line 273: Oh you are doing a query for findUser here, in the two places I commented above, you are doing an extra find user to pass in this user id to this method

oldfashionedrobot (changes requested):

src/app.module.ts
RavenHursT: https://github.com/thegoodparty/gp-api/pull/61/commits/f093631ad009c9b1fc274343b30e292b39873220

src/campaigns/services/campaigns.service.ts
RavenHursT: Whoops! Good catch!

src/campaigns/services/campaigns.service.ts
RavenHursT: https://github.com/thegoodparty/gp-api/pull/61/commits/7c1dcc14fa1cc0b61e1b45200a833910f1701fec

src/fullStory/fullStory.controller.ts
RavenHursT: https://github.com/thegoodparty/gp-api/pull/61/commits/d2209a8f54159963022cf664daa820f302d558b6

src/fullStory/fullStory.controller.ts
RavenHursT: Yeah.. I couldn't think of a better name.. Def didn't wanna do "sewed" as it is in tgp

src/fullStory/fullStory.service.ts
RavenHursT: The `userId` is now coming off the `Campaign` object in those use cases.

oldfashionedrobot (approved):

PR #68, opened 2025-01-28
Web 3436

oakinh:
Adds findBlogArticlesBySlug and modifies findByType to take a prisma where


src/content/content.controller.ts
```diff
@@ -193,4 +195,54 @@ export class ContentController {
...
+        ContentType.blogArticle,
+        undefined,
+        undefined,
+        {
+          data: {
+            path: ['slug'],
+            equals: slug,
+          },
+        },
+      )
```
RavenHursT, line 212: Can we change the signature of this `findByType` to accept an "options" object instead of an ordered list of args so that we can avoid using `undefined` directly in this way? If we find ourselves w/ a method that's accepting more than 2 or 3 args, that's typically a code-smell that either the method is trying to do too much, or that taking an options object would be a more flexible pattern

```suggestion
      await this.contentService.findByType({
        type: ContentType.blogArticle,
        where: {
          data: {
            path: ['slug'],
            equals: slug,
          },
        }
      })
```

RavenHursT (changes requested):
Just 1 thing

oakinh:
https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/38601784-7cab7165-ea89-43c0-a7bc-e8e6a6e88456

RavenHursT (approved):

PR #64, opened 2025-01-27
Prisma base service

oldfashionedrobot:
- updates the rest of the current Prisma model services to use the Prisma base service
- This makes all Prisma `find` methods and `count` automatically available on the child service classes, with all of the existing methods' types
- Makes the specific Prisma model available on the child class's `model` property
- Exposes Prisma client methods via the child class's `client` property

Postman Tests:
- [Admin](https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/38094541-de7cc281-946b-4777-a20e-bd0a5ccf3df6)
- [Campaigns](https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/38094541-f2a3568e-f3c4-4a33-9a24-afa929b23345)
- [CMS Content](https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/38094541-168f5f0b-7d7d-4199-9d7a-9015f143684c)
- [Integrations](https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/38094541-7b169205-b78d-49f0-8c8f-f58e1586e678)
- [Races](https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/38094541-6743607a-8e90-4878-a273-0b14d06e8466)
- [TopIssues](https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/38094541-59cf5916-2c3c-4e3c-a2d7-5a9e8772f109)
- [Users](https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/38094541-2fce364f-ae88-45e1-af27-4821151751e7)

PR #69, opened 2025-01-29
Prisma base service

oldfashionedrobot:
This PR updates the Prisma base service to be a factory style function, and the updates rest of the current Prisma model services to use the Prisma base service. 
- This makes the following Prisma query methods automatically available on the child service classes, utilizing all the type functionality of the prisma method
  - 'findMany',
  - 'findFirst',
  - 'findFirstOrThrow',
  - 'findUnique',
  - 'findUniqueOrThrow',
  - 'count'
- Makes the specific Prisma model available on the child class's model property
- Exposes Prisma client methods via the child class's client property for any client level functons


src/campaigns/ai/chat/aiChat.service.ts
```diff
@@ -133,7 +105,7 @@ export class AiChatService {
       throw new Error('Cannot regenerate without threadId')
     }
 
-    const aiChat = await this.prisma.aiChat.findFirstOrThrow({
+    const aiChat = await this.findFirstOrThrow({
```
oldfashionedrobot, line 108: Note: functions like this are also available on `this.model.findFirstOrThrow` the base service class just also makes these available as public methods here too, so that a consumer of this `AiChatService` can call `findFirstOrThrow` with all the dynamic prisma typings

src/prisma/basePrisma.service.ts
```diff
@@ -1,59 +1,61 @@
...
-    // allows child class to use only their specified model
-    return this._prisma[this.modelName]
-  }
+] satisfies PrismaMethods[]
 
-  onModuleInit() {
-    // make PASSTHROUGH_METHODS directly available on child class as public methods
-    for (const method of PASSTHROUGH_METHODS) {
-      this[method] = this.model[method].bind(this.model)
+export function createPrismaBase<T extends Prisma.ModelName>(modelName: T) {
```
oldfashionedrobot, line 24: changed this to a factory kind of function to simplify use. 
```
class MyCampaignService extends createPrismaBase(MODELS.Campaign) {}
```

src/campaigns/campaigns.controller.ts
```diff
@@ -116,7 +116,7 @@ export class CampaignsController {
       campaign = await this.campaignsService.findFirstOrThrow({
         where: { slug },
       })
-    }
+    } else if (!campaign) throw new NotFoundException('Campaign not found')
```
oldfashionedrobot, line 119: RE: https://github.com/thegoodparty/gp-api/pull/64#discussion_r1931001913

Changed to else if to be more explicit about the case this handles

src/prisma/basePrisma.service.ts
```diff
@@ -1,59 +1,61 @@
...
+  const lowerModelName = lowerFirst(modelName)
+  /* eslint-disable @typescript-eslint/no-unsafe-declaration-merging */
+  @Injectable()
+  class BasePrismaService implements OnModuleInit {
+    @Inject()
+    // NOTE: TS won't let me make this private when returning a class def from a function
+    readonly _prisma!: PrismaService
```
oldfashionedrobot, line 31: RE: https://github.com/thegoodparty/gp-api/pull/64#discussion_r1931006982

Might be a good idea to look into. Though I think the PrismaService pattern lets NestJS handle it as a singleton for us so we don't have to worry about it

src/races/services/counties.services.ts
```diff
@@ -0,0 +1,5 @@
+import { Injectable } from '@nestjs/common'
+import { createPrismaBase, MODELS } from 'src/prisma/basePrisma.service'
+
+@Injectable()
+export class CountiesService extends createPrismaBase(MODELS.County) {}
```
oldfashionedrobot, line 5: Example of bare minimum model service using the base class factory.

RavenHursT (approved):

PR #70, opened 2025-01-30
WEB-3579: Make the web app work with nestjs - Part 0

oldfashionedrobot:
Changes for connecting webapp to new api

- Made more reusable zod schemas for common fields and updated places to use existing reusable schemas for consistency. (email, phone, zip, password, etc)
- updated error messages in the Local auth strategy guard
- added `credentials: true` to cors config


src/authentication/authentication.service.ts
```diff
@@ -106,7 +106,7 @@ export class AuthenticationService {
     )
 
     if (!validPassword) {
-      throw new UnauthorizedException('Invalid password')
+      throw new UnauthorizedException()
```
oldfashionedrobot, line 109: Removed error messages to not allow enumeration of accounts

RavenHursT (approved):

PR #76, opened 2025-02-04
[WEB-3395] - Hubspot Integration

RavenHursT:
ALL the HubSpot things (I hope)

https://goodparty.atlassian.net/browse/WEB-3395

Authentication tests: https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/34024107-c8cd92a4-0315-449f-b3bf-3638031e9f64

Integration tests: https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/34024107-97b1d9e1-667e-4b43-ab64-5e246e287d19


src/campaigns/services/path-to-victory.service/pathToVictory.service.ts
oldfashionedrobot, line 1: Can you move this just into the services folder, and remove the `path-to-victory.service` folder?

src/campaigns/updateHistory/campaignUpdateHistory.service.ts
```diff
@@ -8,7 +8,10 @@ import { createPrismaBase, MODELS } from 'src/prisma/util/prisma.util'
 export class CampaignUpdateHistoryService extends createPrismaBase(
   MODELS.CampaignUpdateHistory,
 ) {
-  constructor(private readonly campaigns: CampaignsService) {
+  constructor(
+    @Inject(forwardRef(() => CampaignsService))
+    private readonly campaigns: CampaignsService,
```
oldfashionedrobot, line 13: Do you need the forward ref here?, I don't see the campaignsService injecting this service

src/campaigns/campaigns.module.ts
```diff
@@ -12,9 +12,21 @@ import { GeocodingService } from './services/geocoding.service'
...
+import { ContentModule } from '../content/content.module'
+import { CrmModule } from '../crm/crmModule'
 
 @Module({
-  imports: [EmailModule, forwardRef(() => UsersModule), RacesModule],
+  imports: [
+    EmailModule,
+    forwardRef(() => UsersModule),
+    RacesModule,
+    AiModule,
```
oldfashionedrobot, line 26: Does the AiModule and ContentModule need to be imported here?

src/campaigns/campaigns.module.ts
```diff
@@ -28,7 +40,14 @@ import { CampaignUpdateHistoryService } from './updateHistory/campaignUpdateHist
     CampaignMapService,
     GeocodingService,
     CampaignUpdateHistoryService,
+    PathToVictoryService,
+    AiChatService,
```
oldfashionedrobot, line 44: Does the AiChatService need to be here? Seems to be misplaced

src/voters/voters.module.ts
```diff
@@ -7,9 +7,15 @@ import { VoterOutreachService } from './services/voterOutreach.service'
 import { FilesModule } from 'src/files/files.module'
 import { VotersService } from './services/voters.service'
 import { HttpModule } from '@nestjs/axios'
+import { CrmModule } from '../crm/crmModule'
 
 @Module({
-  imports: [CampaignsModule, FilesModule, HttpModule],
+  imports: [
+    forwardRef(() => CampaignsModule),
```
oldfashionedrobot, line 14: I don't see the CampaignsModule importing the VotersModule, I don't think you need this forwardRef here.

src/crm/crmCampaigns.service.ts
```diff
@@ -0,0 +1,706 @@
...
+    private readonly pathToVictory: PathToVictoryService,
+    private readonly campaignUpdateHistory: CampaignUpdateHistoryService,
+    private readonly voterFile: VoterFileService,
+    @Inject(forwardRef(() => CampaignsService))
+    private readonly campaigns: CampaignsService,
+    private readonly slack: SlackService,
+    @Inject(forwardRef(() => UsersService))
+    private readonly users: UsersService,
+    @Inject(forwardRef(() => FullStoryService))
+    private readonly fullStory: FullStoryService,
```
oldfashionedrobot, line 83: This service should really be broken down, not necessarily in this PR, but I think we are defeating the purpose of using NestJS if we have to inject the whole app into a service. 

(I know this is a result of the way the code was in tgp-api)

oldfashionedrobot: Maybe this is a situation to just inject and use the PrismaService/PrismaClient directly. 

It needs to query so many different models, and kind of breaks the modularizing patterns we are trying to build in NestJS. 

There are so many circuluar dependencies, that usually means this needs to be reorganized , but maybe we do want to have our CRM tracking module to be able to query whatever it needs without having to reorganize our modules just for this

src/crm/crmModule.ts
```diff
@@ -0,0 +1,22 @@
...
+import { FullStoryModule } from '../fullStory/fullStory.module'
+
+@Module({
+  providers: [CrmUsersService, CrmCampaignsService, HubspotService],
+  imports: [
+    forwardRef(() => CampaignsModule),
+    forwardRef(() => VotersModule),
+    forwardRef(() => UsersModule),
+    HttpModule,
+    forwardRef(() => FullStoryModule),
```
oldfashionedrobot, line 18: Do all these forwardRefs need to be here? I think if every dependency is a circular dependency, things need to be reorganized

oldfashionedrobot (changes requested):
I think maybe this needs to be reorganized, the crmUsers service for example, should be in a CRMUsers module inside the UsersModule, which imports the CRM module. 

But maybe the NestJS pattern is a hindrance here, and we should just let the CRM module have access to query Prisma directly. I think it makes sense in this case, since we need to pull data from potentially anywhere in the DB to send to Hubspot. 

Otherwise, we'd have to break this thing down to avoid all the circular dependencies and maybe that's more work than it's worth, when really we want a service that can query whatever it needs.

src/crm/crmCampaigns.service.ts
```diff
@@ -0,0 +1,706 @@
...
+  'jared@goodparty.org': 'U01AY0VQFPE',
+  'ryan@goodparty.org': 'U06T7RGGHEZ',
+  'kyron.banks@goodparty.org': 'U07JWLYDDUH',
+  'alex.barrio@goodparty.org': 'U0748BRPPJQ',
+  'trey.stradling@goodparty.org': 'U06FPEP4QBZ',
+  'alex.gibson@goodparty.org': 'U079ASLQ9G8',
+  'dllane2012@gmail.com': 'U06U033GHDE',
+  'aaron.soriano@goodparty.org': 'U07QXHVNDEJ',
+  'nate.allen@goodparty.org': 'U07R9RNFTFX',
+}
```
oakinh, line 65: We need some kind of semi-automated way to make sure we're updating this consistently as PA's are onboarded and offboarded

prisma/schema/campaign.jsonTypes.ts
```diff
@@ -66,14 +68,18 @@ declare global {
         verified_candidates?: string
         election_results?: string
         office_type?: string
-      }
+      } & Record<string, string>
```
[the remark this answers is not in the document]
RavenHursT: The reason we have to have this is because the webhook endpoint takes the "firehose" of each company and then persists it's properties into this json column.

In order to take this `Record<string, string>` off of here, we'd have to somehow compile a list of each and ever property that HubSpot is configured for Companies in our account rn. Not to mention, at any given moment, users in HS can add and modify/rename company properties at will.  

Until we have an actual, comprehensive integration w/ HS, I'm not sure how else we're supposed to set up this type 🤔

src/app.module.ts
```diff
@@ -42,9 +44,10 @@ import { SubscribeModule } from './subscribe/subscribe.module'
     PaymentsModule,
     VotersModule,
     ErrorLoggerModule,
+    CrmModule,
     SubscribeModule,
   ],
-  controllers: [AppController],
+  controllers: [AppController, CrmController],
```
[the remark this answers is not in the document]
RavenHursT: https://github.com/thegoodparty/gp-api/pull/76/commits/959e696731c9c0720707c9efc536abb8192f7c4b

src/campaigns/services/path-to-victory.service/pathToVictory.service.ts
RavenHursT: https://github.com/thegoodparty/gp-api/pull/76/commits/2021b49739eda8556d8536535c0a263ba5c633c6

src/campaigns/updateHistory/campaignUpdateHistory.service.ts
RavenHursT: Yeah.. as soon as I take it out.. I get dependency errors 

image

src/campaigns/campaigns.module.ts
RavenHursT: I mean.. it's in the `/src/campaigns` module folder, so I'm guessing it's supposed to be? 🤷 

https://github.com/thegoodparty/gp-api/blob/9483f4a7cadd3df80b5415120f35c7c4e69188d8/src/campaigns/ai/chat/aiChat.service.ts#L17-L16

src/crm/crmModule.ts
RavenHursT: I'll try to do your idea about moving the users crm service under `Users` and `Campaigns` for the other.. Hopefully that'll untangle this nonsense. Spent hours on this crap.

package.json
```diff
@@ -77,6 +77,7 @@
     "sanitize-html": "^2.14.0",
     "slugify": "^1.6.6",
     "stripe": "^17.5.0",
+    "us-state-codes": "^1.1.2",
```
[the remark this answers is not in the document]
RavenHursT: https://github.com/thegoodparty/gp-api/pull/76/commits/53f9f28644dbbcaad5617d7f220720b2b10005a8

src/campaigns/updateHistory/campaignUpdateHistory.service.ts
oldfashionedrobot: I think we should find out why instead of throwing forwardRef's everywhere. We chose an opininated framework, we kind of should try to use the opinions to organize stuff in a way that doesn't have circular dependencies for so many things

src/campaigns/campaigns.module.ts
oldfashionedrobot: There's a campaignsAi.module inside there too that has the campaign Ai services registered there. I don't think the AiChatService needs to be here

src/crm/crmCampaigns.service.ts
RavenHursT: I don't agree. I really think abandoning a service-layer pattern will come back to bite us later.

If we have any kind of business logic in our services for reading or writing data, we need a place for those to be in _one_ place for each model.  If we abandon that and start making one offs, we end up in the same place we were in the Sails app, having to remember to apply that business logic in different places in the app and inevitably introducing bugs when we forget to do so.

I personally don't have a problem w/ the circular `forwardRef`s.  Apart from having to have them in the first place.  I see it more as a flaw in the Angular-mimiked dependency injection model Nest uses, than an architecture problem for our app. I haven't read anything suggesting that utilizing `forwardRef` for modules has some kind of negative impact on app performance (maybe I'm missing that?), so I'm sure what harm there is in using it.  Nest provides it for a reason.

Regardless, I do agree that this service is a mess, and we def need to clean up this code that was a mess in the Sails app before porting.  In fact, pretty much all of this Races code is rather hard to grok and navigate and could use some love.

But for now, I'm gonna leave as-is so we can just get this to feature-parity and get it out the door.

I'm also going to at least try out your idea w/r/t moving this service into the `CampaignsModule` and the other into `UsersModule` to see if it lessens the # of `forwardRef`s that we have to use at least.

oldfashionedrobot: I'm cool with that for now. Though while nest gives us forwardRef, they recommend avoiding using it when possible: https://docs.nestjs.com/fundamentals/circular-dependency

It's there to let you use a circular dependency when there is no other choice. And if we are having to use it because we are creating circular dependencies and can't identify them, then there is an issue in how we are organizing things. 

But also, I agree, we don't have to figure it out this moment

RavenHursT: https://github.com/thegoodparty/gp-api/pull/76/commits/3fbc04dd5bf0cecf088e4292803a5e3fcd3dd82c

Well.. it limited a _few_ `forwardRef`s, but had to also add one or two.. so 🤷 

But I split it out for ya..

RavenHursT: One more: https://github.com/thegoodparty/gp-api/pull/76/commits/0562dd6cbefc77de53f38fb99dd171b1c01fcf88

src/campaigns/campaigns.module.ts
RavenHursT: https://github.com/thegoodparty/gp-api/pull/76/commits/89d7067fef1bb1e36dd35c50306c031d5cb729f4

src/crm/crmModule.ts
RavenHursT: https://github.com/thegoodparty/gp-api/pull/76/commits/b7d3bab218abdde05035274e04170f2cd8482c59

seed/campaigns.ts
```diff
@@ -9,14 +9,15 @@ import { campaignUpdateHistoryFactory } from './factories/campaignUpdateHistory.
...
+import { getUserFullName } from 'src/users/util/users.util'
 import { Campaign } from '@prisma/client'
 import fixedCampaigns from './fixedCampaigns.json'
 import 'prisma/schema/pathToVictory.jsonTypes'
 import { campaignPlanVersionFactory } from './factories/campaignPlanVersion.factory'
 const NUM_GENERATED_CAMPAIGNS = 100
 const NUM_UPDATE_HISTORY = 3
-const FIXED_CAMPAIGNS: Partial<Campaign>[] = fixedCampaigns
+const FIXED_CAMPAIGNS: Partial<Campaign>[] =
+  fixedCampaigns as Partial<Campaign>[]
```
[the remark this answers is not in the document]
RavenHursT: Yeah...

image

src/crm/crm.types.ts
```diff
@@ -0,0 +1,88 @@
...
+  doors_knocked?: string
+  direct_mail_sent?: string
+  calls_made?: string
+  online_impressions?: string
+  p2p_sent?: string
+  event_impressions?: string
+  yard_signs_impressions?: string
+  my_content_pieces_created?: string
+  filed_candidate?: 'yes' | 'no'
+  pro_candidate?: 'Yes' | 'No'
```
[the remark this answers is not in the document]
RavenHursT: These are the values set by the Marketing team in HubSpot.  I agree.. it's frustrating.. but HS will choke if we send anything else across the wire... 😡

src/crm/crmCampaigns.service.ts
```diff
@@ -0,0 +1,706 @@
...
+      ...(typeof isPartisan === 'boolean'
+        ? { partisan_np: isPartisan ? 'Partisan' : 'Nonpartisan' }
+        : {}),
+    }
+
+    delete properties.winnumber
+    delete properties.p2vStatus
+    delete properties.p2vstatus
+    delete properties.p2vCompleteDate
+    delete properties.p2vcompletedate
```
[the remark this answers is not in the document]
RavenHursT: I think they were in there because there were fat-fingered properties in HS that were getting injected from the webhook and then would fail when we try to send them back over.

These are in the original logic over in tgp, so for now we're leaving them for safety.. we really need to _totally_ rip all this out and rewrite the HS integration.

src/crm/crmCampaigns.service.ts
```diff
@@ -0,0 +1,706 @@
...
+      my_content_pieces_created: `${aiContent ? Object.keys(aiContent).length : 0}`,
+      filed_candidate: campaignCommittee ? 'yes' : 'no',
+      pro_candidate: isPro ? 'Yes' : 'No',
+      pro_upgrade_date: isProUpdatedAtMs,
+      filing_start: filingStartMs,
+      filing_end: filingEndMs,
+      ...(website ? { website } : {}),
+      ...(level ? { ai_office_level: level } : {}),
+      ...(ballotLevel ? { office_level: ballotLevel } : {}),
+      ...(runForOffice ? { running: runForOffice ? 'yes' : 'no' } : {}),
```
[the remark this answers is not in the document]
RavenHursT: Good catch

https://github.com/thegoodparty/gp-api/pull/76/commits/699ff6e6960821cf3ba6c7c9e1fb4f16a8668c6c

src/crm/crmCampaigns.service.ts
```diff
@@ -0,0 +1,706 @@
...
+        : {}),
+      ...(typeof isIncumbent === 'boolean'
+        ? { incumbent: isIncumbent ? 'Yes' : 'No' }
+        : {}),
+      ...(seats && typeof seats === 'number' && seats > 0
+        ? { seats_available: `${seats}` }
+        : {}),
+      ...(score && typeof score === 'number' && score > 0
+        ? { automated_score: `${Math.floor(score > 5 ? 5 : score)}` }
+        : {}),
```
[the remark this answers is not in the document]
RavenHursT: https://github.com/thegoodparty/gp-api/pull/76/commits/f8452e229c2d4c8079099bfdbe6faa645396326b

src/crm/crmCampaigns.service.ts
```diff
@@ -0,0 +1,706 @@
...
+        const company = hubspotId
+          ? await this.getCrmCompanyById(hubspotId)
+          : null
+        if (!company) {
+          this.logger.error(`No company found - ${campaignId}`)
+          continue
+        }
+
+        // TODO: verify this does nothing and then kill it
+        // sleep(400)
```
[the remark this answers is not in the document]
RavenHursT: https://github.com/thegoodparty/gp-api/pull/76/commits/2e23ec54a6bcb83030fa52f3311aa081ee5c581e

src/users/util/users.util.ts
```diff
@@ -1,8 +1,13 @@
 import { User, UserRole } from '@prisma/client'
 
-export function getFullName(user: User) {
-  return `${user.firstName} ${user.lastName}`
-}
+export const getUserFullName = (user: User) =>
+  !user
+    ? ''
+    : user.firstName
+      ? `${user.firstName} ${user.lastName}`
```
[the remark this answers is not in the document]
RavenHursT: This is prolly even better: https://github.com/thegoodparty/gp-api/pull/76/commits/8e3440d7ba56153b7937e486a896c378d50899ea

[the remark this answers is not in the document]
oakinh: I think using trim() is a little cleaner but this is good too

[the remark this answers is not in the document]
RavenHursT: So hard to please... 

https://github.com/thegoodparty/gp-api/pull/76/commits/a0e530aa148dd075403bbc95415e1e68ff824879

oldfashionedrobot (approved):

PR #78, opened 2025-02-06
Ravenhurst/web 3389 web 3390 races listing endpoints

RavenHursT:
https://goodparty.atlassian.net/browse/WEB-3389 and https://goodparty.atlassian.net/browse/WEB-3390

Postman run: https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/34024107-2dc2128d-c256-4c35-ac03-b9bf5953ccf9

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,8 @@
     "axios": "^1.7.7",
     "bcrypt": "^5.1.1",
     "bottleneck": "^2.19.5",
+    "class-transformer": "^0.5.1",
+    "class-validator": "^0.14.1",
     "contentful": "^11.2.2",
     "date-fns": "^4.1.0",
     "deepmerge-ts": "^7.1.3",
--- a/package.json
+++ b/package.json
@@ -52,8 +52,6 @@
     "axios": "^1.7.7",
     "bcrypt": "^5.1.1",
     "bottleneck": "^2.19.5",
-    "class-transformer": "^0.5.1",
-    "class-validator": "^0.14.1",
     "contentful": "^11.2.2",
     "date-fns": "^4.1.0",
     "deepmerge-ts": "^7.1.3",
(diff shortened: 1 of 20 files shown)


package.json
```diff
@@ -52,6 +52,8 @@
     "axios": "^1.7.7",
     "bcrypt": "^5.1.1",
     "bottleneck": "^2.19.5",
+    "class-transformer": "^0.5.1",
+    "class-validator": "^0.14.1",
```
oldfashionedrobot, line 56: Do we need these packages? we are using zod for validations, we should also need to be using these.

edit: i dont think I see these being used anywhere

src/main.ts
```diff
@@ -71,6 +71,7 @@ const bootstrap = async () => {
     },
   })
 
+  app.useGlobalPipes(new ValidationPipe({ transform: true }))
```
oldfashionedrobot, line 74: Do we need this? We could add a ZodValidationPipe here globally and not have to add one in controllers.

oldfashionedrobot (changes requested):
Two small things, all looks good otherwise

src/main.ts
RavenHursT: No we don't need it. I forgot to take it out when I was trying to figure out why validation wasn't working for the controller that I was working on (I hadn't added the pipe to that controller).

package.json
RavenHursT: Another side effect of my trying to get validation working for that one controller.

RavenHursT: https://github.com/thegoodparty/gp-api/pull/78/commits/948b27f338ceeae02314bced4fb74a542da20275

oldfashionedrobot (approved):

PR #80, opened 2025-02-06
Web 3459

oakinh:
The massive files of types under generated/ is generated from the BR/CivicEngine GQL Schema. "codegen" script in package.json has been added to generate this. 

The races tests are broken in dev, so I wasn't able to verify all of those endpoints, but the BallotData endpoint postman test is found below.

https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/request/38601784-1a360b88-a5a8-476a-bfc7-49a00bf22a97?action=share&source=copy-link&creator=38601784&active-environment=0205933f-8975-40bb-89f4-1c3db95f27e7


src/email/mailgun.service.ts
```diff
@@ -5,6 +5,9 @@ import { IMailgunClient } from 'mailgun.js/Interfaces'
 
 const EMAIL_DOMAIN = 'mg.goodparty.org'
 const API_KEY = process.env.MAILGUN_API_KEY as string
+if (!API_KEY) {
+  throw new Error('Please set MAILGUN_API_KEY in your .env')
```
oldfashionedrobot, line 9: I don't know if we should throw errors for config vars that aren't app critical. In theory the app should run fine without having a mailgun setup. 

Maybe for non-critical vars, you can do a logger.error or something similar so that app can continue, but a very visible message still shows for devs

src/races/races.controller.ts
```diff
@@ -14,7 +18,10 @@ import { ZodValidationPipe } from 'nestjs-zod'
 @UsePipes(ZodValidationPipe)
 @PublicAccess()
 export class RacesController {
-  constructor(private readonly racesService: RacesService) {}
+  constructor(
+    private readonly racesService: RacesService,
+    private readonly ballotDataService: BallotDataService,
```
oldfashionedrobot, line 23: is `ballotDataService` used in this controller?

src/races/services/ballotReadyservice.ts
oldfashionedrobot, line 1: Looks like you missed a dot before `service` in the filename

src/races/services/races.service.ts
```diff
@@ -828,37 +829,15 @@ export class RacesService extends createPrismaBase(MODELS.Race) {
...
-                    }
-                }
-            }`
-
-      const { races } = await this.graphQLService.fetchGraphql(query)
+      const { races } =
+        (await this.ballotReadyService.fetchRacesWithElectionDates(
+          zip,
+          level,
+        )) as any
```
oldfashionedrobot, line 840: Do you need this `any` here? Shouldn't the type be `RacesWithElectionDates`? Or at least if you need to cast this type here you can cast it to an object with a `races` property that you expect here.

oldfashionedrobot (changes requested):

src/email/mailgun.service.ts
oakinh: I get an error on startup if the key isn't set, that's why I added it

src/races/services/races.service.ts
oakinh: Good catch, I forgot to change this and turned out I had a bug with extra wrapping of a field in the types. Thank you!

src/races/races.controller.ts
oakinh: Ah good catch, fixed

RavenHursT (changes requested):

oldfashionedrobot (approved):
lgtm after matthews comments

src/races/services/races.service.ts
```diff
@@ -828,37 +829,15 @@ export class RacesService extends createPrismaBase(MODELS.Race) {
...
       this.logger.log(slug, 'getElectionDates graphql result', races)
```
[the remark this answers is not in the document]
oakinh: There's a total of 17 this.logger.log's in the races.service file. Can I change them all to this.logger.debug's or just this one?

PR #79, opened 2025-02-06
WEB-3605: More fixes for webapp 

oldfashionedrobot:
- added voter locations endpoint
- updated voter routes to be under `voters` instead of `voter-data`
- fixes some content transformers from array of one item to just return the item 

Postman tests:
- [Voters](https://goodpartyorg.postman.co/workspace/66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/38094541-79425f4c-3f2b-4a69-9a32-09f486a45281) (my fork) - [Postman PR](https://goodpartyorg.postman.co/workspace/66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/pull-request/d4710ef1-906a-4bc1-85da-19dd837890fe)
- [Content](https://goodpartyorg.postman.co/workspace/66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/38094541-4ea44d9c-6e90-4ca2-8a48-50c3caf735f8)

PR #82, opened 2025-02-10
WEB-3606: fixes for race endpoints

oldfashionedrobot:
**NOTE: based on #79**

Postman tests:
- [Races](https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/38094541-a3535eab-bb58-4147-ba1c-bb7a3b297e3e)


src/races/services/races.service.ts
```diff
@@ -87,35 +83,74 @@ export class RacesService extends createPrismaBase(MODELS.Race) {
...
       return null
     }
 
-    race.municipality = cityRecord
-    race.county = countyRecord
+    const race = races[0]
+    const positions: Array<string | undefined> = []
+    for (let i = 0; i < races.length; i++) {
+      positions.push(races[i].data.position_name)
+    }
```
RavenHursT, line 119: Guessing this is just copy/pasted from tgp.. but this _could_ just be a `.map()` if we wanna do a bit of cleanup.

src/races/services/races.service.ts
```diff
@@ -87,35 +83,74 @@ export class RacesService extends createPrismaBase(MODELS.Race) {
...
+    let otherRaces: Race[] = []
+    if (race.municipality) {
+      otherRaces = await this.findMany({
+        where: { municipalityId: race.municipality.id },
+      })
+    } else if (race.county) {
+      otherRaces = await this.findMany({
+        where: { countyId: race.county.id },
+      })
+    }
```
RavenHursT, line 130: this could also be a simple `const` w/ a ternary

RavenHursT (approved):
Just a couple comments on some possible clean up

PR #87, opened 2025-02-13
WE-3441: Migrated Mass CRM refresh, CSV file seeds

oldfashionedrobot:
- Adds the `mass-refresh-companies` endpoint for batch refreshing hubspot records
- Migrates some CSV file seeds, and related CSV files they pull from
  - mtfcc
  - election types
  - counties
  - municipalites


seed/municipalities.ts
```diff
@@ -0,0 +1,123 @@
...
+    },
+    update: {},
+    create: {
+      slug,
+      name: county_name,
+      state: state_id,
+      data: row,
+    },
+  })
+}
```
RavenHursT, line 123: Should this not be in `seed/counties.ts`?

seed/seed.ts
```diff
@@ -1,16 +1,50 @@
...
 
+// csv file seeds
+import seedElectionTypes from './electionTypes'
+import seedMtfcc from './mtfcc'
+import seedCounties from './counties'
+import seedMunicipalities from './municipalities'
+
+const IS_PROD_BUILD =
+  process.env.NODE_ENV === 'production' || process.env.NODE_ENV === 'qa'
+
```
RavenHursT, line 18: I think we need to add `development` here as well so that the seeds don't wipe-out that DB as well

src/crm/crm.controller.ts
```diff
@@ -77,15 +86,22 @@ export class CrmController {
...
-  async refreshCompanies(@Query('campaignId') campaignId: number) {
+  async refreshCompanies(@Query() { campaignId }: RefreshCompanySchema) {
     return await this.crmCampaignsService.refreshCompanies(campaignId)
   }
 
+  @Get('mass-refresh-companies')
+  @Roles(UserRole.admin)
+  async massRefreshCompanies(@Query() { fields }: MassRefreshCompanySchema) {
+    return await this.crmCampaignsService.massRefreshCompanies(
+      fields as Array<keyof CRMCompanyProperties>,
```
RavenHursT, line 97: We wouldn't need to do this type-casting if we use the keys from `CRMCompanyProperties` for the Zod `MassRefreshCompanySchema.fields` instead of just `z.array(z.string())`, right?

RavenHursT (changes requested):

seed/seed.ts
oldfashionedrobot: I'll just check if not local instead

src/crm/crm.controller.ts
oldfashionedrobot: I was looking into a good way to do that, we'd have to write a zod schema with all those keys, then infer the type from that. There's not a good way to pull the keys from that type to use at runtime. 

It doesn't error if the keys don't match any fields, it just skips them.

seed/seed.ts
RavenHursT: Don't we need all the seeds to run on branch-preview envs tho? :thinking_face:

So yeah.. in prod, qa, and dev, we only want the "csv" seeds to run.. everywhere else, do it all..

RavenHursT (approved):

PR #96, opened 2025-02-19
feat: WEB-3643 new office picker

tomer-tgp:


oldfashionedrobot (approved):

src/elections/services/ballotReady.service.ts
```diff
@@ -115,12 +115,28 @@ export class BallotReadyService {
...
+      nextYear.setFullYear(nextYear.getFullYear() + 2)
+      lt = nextYear.toISOString().split('T')[0]
+    }
+
+    let levelWithTownship = level?.toUpperCase()
+    if (levelWithTownship === 'LOCAL') {
+      levelWithTownship = 'LOCAL,TOWNSHIP,CITY'
+    }
+    if (levelWithTownship === 'COUNTY') {
+      levelWithTownship = 'COUNTY,REGIONAL'
```
RavenHursT, line 138: Can we please reference the `const` we have exported here? 

https://github.com/thegoodparty/gp-api/blob/8e0369362cb9c48526b3366e7c09dca347770f63/src/shared/constants/governmentLevels.ts#L1

src/elections/services/races.service.ts
```diff
@@ -38,29 +37,37 @@ export class RacesService {
     return this.ballotReadyService.fetchRaceNormalizedPosition(raceId)
   }
 
-  async racesByYear({
+  async getRaces({
```
RavenHursT, line 40: Can we rename this `getRacesByZip` to reflect the required input `zip` argument instead of this sounding like a generic "get any races" method?

RavenHursT (changes requested):

RavenHursT (approved):

PR #99, opened 2025-02-20
[WEB-3367] - Adds missing setTokenCookie

RavenHursT:


src/authentication/authentication.controller.ts
```diff
@@ -85,13 +79,19 @@ export class AuthenticationController {
...
+  async socialLogin(
+    @Res({ passthrough: true }) response,
+    @ReqUser() user: User,
+  ): Promise<LoginResult> {
+    const token = this.authenticationService.generateAuthToken({
+      email: user.email,
+      sub: user.id,
+    })
+
+    setTokenCookie(response, token)
```
oldfashionedrobot, line 91: Why did we need this set token cookie here? Won't it be for the backend api's domain, not the frontend app domain?

RavenHursT: I was having issues getting my user lookup to work through the UI.. and I suspect it was because of the `/api/set-cookie` hack we have in the UI not working for some reason.

Regardless, the headers from the API response are reverse-proxied back up through the `/api/*` path in the NextJS app so the headers will be apply to the client page url (This is on a local production build as well):

Screenshot 2025-02-20 at 3 03 18 PM
Screenshot 2025-02-20 at 3 06 19 PM

PR #102, opened 2025-02-25
feat: return 498 error code on invalid token

tomer-tgp:


src/authentication/guards/JwtAuth.guard.ts
```diff
@@ -12,21 +17,48 @@ export class JwtAuthGuard extends AuthGuard('jwt') {
...
+    // Skip JWT authentication if the route is public and does not have role restrictions
+    if (isPublic && !roles) {
+      return true
+    }
 
+    // Otherwise, require JWT authentication
     return super.canActivate(context)
   }
+
+  handleRequest(err: any, user: any, info: any) {
```
RavenHursT, line 41: Can we not type these?

`user` can at least be typed after the `User` model, no? 🤔

src/authentication/guards/JwtAuth.guard.ts
```diff
@@ -12,21 +17,48 @@ export class JwtAuthGuard extends AuthGuard('jwt') {
...
+      if (
+        info &&
+        (info.name === 'JsonWebTokenError' || info.name === 'TokenExpiredError')
+      ) {
+        throw new HttpException(
+          {
+            statusCode: 498,
+            message: 'Invalid or expired token',
+          },
+          498,
```
RavenHursT, line 54: That's annoying that `HttpStatus` doesn't have a constant for this. Can we at least add a comment saying something like:

```suggestion
          498, // Invalid Token Status
```

src/authentication/guards/JwtAuth.guard.ts
```diff
@@ -12,21 +17,48 @@ export class JwtAuthGuard extends AuthGuard('jwt') {
...
+        )
+      }
```
RavenHursT, line 56: Looks like the idiomatic, recommended way to handle a custom exception, not included in the `@nest/common` package, is to create a "custom exception" that extends `HttpException`.

Might be worth looking at for this case?

https://github.com/nestjs/nest/issues/6657#issuecomment-799214959

RavenHursT (changes requested):

src/authentication/guards/JwtAuth.guard.ts
RavenHursT: NestJS Custom Http Exceptions docs: https://docs.nestjs.com/exception-filters#custom-exceptions

src/authentication/guards/JwtAuth.guard.ts
```diff
@@ -1,32 +1,65 @@
...
+    // Otherwise, require JWT authentication
     return super.canActivate(context)
   }
+
+  handleRequest<TUser = any>(
+    err: any,
+    user: TUser,
+    info: any,
+    context: ExecutionContext,
+  ): TUser {
```
RavenHursT, line 45: @tomer-tgp https://github.com/thegoodparty/gp-api/pull/103/files

RavenHursT (changes requested):

RavenHursT (approved):

PR #106, opened 2025-02-26
Web 3721 - Fix campaign map queries

oakinh:
Correct campaign map methods to query correctly. This will now return the 3000+ results we are expecting


hhkarimi (approved):

src/campaigns/map/campaignMap.service.ts
```diff
@@ -57,7 +58,6 @@ export class CampaignMapService {
     const where: Prisma.CampaignWhereInput = {
       userId: { not: undefined },
       isDemo: false,
-      isActive: true,
```
tomer-tgp, line 60: Why did we remove this?

RavenHursT (approved):

src/campaigns/map/campaignMap.service.ts
oakinh: Well as far as I can tell this isn't done on tgp-api at all. So I'm thinking this got added in by AI being dumb, this was one of the few things I let AI do, and we can see how well that went...

RavenHursT (approved):

src/campaigns/map/campaignMap.service.ts
RavenHursT: @tomer-tgp ?

tomer-tgp: we want to show only active campaigns. non active means we removed a user from our site for some reason, and we shouldn't show them on the map.
did we keep this attribute on the campaign?

RavenHursT: It's on the Prisma model 🤷‍♂️ 

https://github.com/thegoodparty/gp-api/blob/0099c287b3c557ea62e127e35fbc223873883703/prisma/schema/campaign.prisma#L12

oakinh: Ahhh I see, I'll change it back!

PR #108, opened 2025-02-26
WEB-3718

oakinh:
Update webAppRootUrl for develop to be the 3367 feature branch


taylorgp (commented):
i think you need to remove the / at the end of the url

oakinh:
> i think you need to remove the / at the end of the url

Ah yep duh, my bad. Fixed.

PR #116, opened 2025-02-28
WEB-3756 & WEB-3763

oakinh:
- Update getAiContentPrompts() to query for the right data with the service method through the transformers instead of directly to prisma
- Remove bug where FS would run when disabled
Frontend: https://github.com/thegoodparty/gp-webapp/pull/493

### DIFF
--- a/src/fullStory/fullStory.service.ts
+++ b/src/fullStory/fullStory.service.ts
@@ -201,7 +201,6 @@ export class FullStoryService {
   }
 
   async trackEvent(user: User, eventName: string, properties: any) {
-    this.logger.debug(`this.disabled: ${this.disabled}`)
     if (this.disabled) {
       this.logger.warn(`FullStory is disabled`)
       return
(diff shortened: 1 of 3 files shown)


src/fullStory/fullStory.service.ts
```diff
@@ -201,6 +201,11 @@ export class FullStoryService {
   }
 
   async trackEvent(user: User, eventName: string, properties: any) {
+    this.logger.debug(`this.disabled: ${this.disabled}`)
+    if (this.disabled) {
+      this.logger.warn(`FullStory is disabled`)
+      return
+    }
```
RavenHursT, line 208: I don't think we need _both_ of these, no?

If we get the warning `Fullstory is disabled`, then `this.disabled === false`, no?  No need for a log for that, right?

RavenHursT (changes requested):

src/fullStory/fullStory.service.ts
oakinh: Yeah I don't think so, but it's how it's done on the other methods. I'll go ahead and remove it

RavenHursT (approved):

PR #117, opened 2025-02-28
WEB-3721: updated canddiate map PROD filter to include is_verified

oldfashionedrobot:
- adds an OR to the prod only filter that checks the hubspotUpdates for 'verified_candidates', to also check the `is_verified` field


oakinh:
If you're removing the userId not undefined condition you'll need to do it for the count method as well

RavenHursT (approved):

oakinh (approved):

PR #121, opened 2025-03-01
fix: ai content updates.

taylorgp:


src/campaigns/services/campaigns.service.ts
```diff
@@ -120,37 +120,60 @@ export class CampaignsService extends createPrismaBase(MODELS.Campaign) {
...
-                },
-              },
-            }
-          : {}),
+      // Track campaign and user
+      this.crm.trackCampaign(campaign.id)
+      campaign.userId && this.usersService.trackUserById(campaign.userId)
+
+      // Handle data and details JSON fields
+      const campaignUpdateData: any = {}
```
oldfashionedrobot, line 128: Instead of any you can pull what you need off the campaign type
```suggestion
      const campaignUpdateData: Pick<Campaign, 'data' | 'details' | 'aiContent'>
```

oldfashionedrobot (changes requested):
Lgtm, just one typescript thing

src/campaigns/services/campaigns.service.ts
taylorgp: that doesn't work. i cant get it to work without the any. you are welcome to take a crack at it. if you do so please make sure you test it.

RavenHursT: https://github.com/thegoodparty/gp-api/pull/123

taylorgp: i merged in the fix it and tested it it should be good to merge now.

PR #122, opened 2025-03-01
WEB-3772: [NestJS] Candidate map cannot be filtered by level

oldfashionedrobot:
- fixes campaign map filtering by level
- upgrades to prisma 6.4.0 to use the new case insensitive mode for json field filtering
- Also fixed level filtering on the admin campaign list page

[Postman Campaigns and Campaigns/Map runs](https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/38094541-9d29aa96-8c0d-4127-833a-108e9f859721)


package.json
```diff
@@ -118,7 +118,7 @@
     "eslint-plugin-unused-imports": "^4.1.4",
     "jest": "^29.5.0",
     "prettier": "^3.0.0",
-    "prisma": "^6.3.0",
+    "prisma": "^6.4.0",
```
RavenHursT, line 121: Only thing I'd say we could consider is doing this update after launch, just in case there's some unforeseen side-effects with it, being this close to launch and all...

RavenHursT (approved):

package.json
oldfashionedrobot: I'll double/triple check the changes before I merge

PR #130, opened 2025-03-04
Ecanvaseer poc

tomer-tgp:


prisma/schema/ecanvasser.prisma
```diff
@@ -0,0 +1,263 @@
...
+    email        String?
+    phoneNumber  String?    @map("phone_number")
+    countryCode  String?    @map("country_code")
+    joined       DateTime
+    billing      Boolean    @default(false)
+    ecanvasser   Ecanvasser @relation(fields: [ecanvasserId], references: [id], onDelete: Cascade)
+    ecanvasserId Int        @map("ecanvasser_id")
+
+    @@index([ecanvasserId])
+    @@map("ecanvasser_users")
```
oldfashionedrobot, line 262: Minor issue, but we have been using singular case for our tables in the Nest project

prisma/schema/ecanvasser.jsonTypes.ts
```diff
@@ -0,0 +1,280 @@
+export {}
+
+declare global {
+  export namespace PrismaJson {
+    export type EcanvasserAppointments = Array<{
```
oldfashionedrobot, line 5: Do we need these types? I see we have prisma models for each of these, so prisma will generate types for those. 

These .jsonTypes files are for annotating a JSON field on a model with a specific type

src/ecanvasser/ecanvasser.types.ts
```diff
@@ -0,0 +1,236 @@
...
+    | 'Meeting'
+  contact_id: number
+  interaction_id?: number | null
+  assigned_to?: number | null
+  created_by: number
+  created_at: string
+  updated_at: string
+}
+
+export interface EcanvasserHouse {
```
oldfashionedrobot, line 124: Prisma should be generating these types, unless these are different than what the prisma models are

oldfashionedrobot (changes requested):

src/ecanvasser/ecanvasser.types.ts
tomer-tgp: removed

prisma/schema/ecanvasser.jsonTypes.ts
tomer-tgp: removed

oldfashionedrobot (approved):

PR #136, opened 2025-03-06
feat: update viability score.

taylorgp:


src/pathToVictory/services/viability.service.ts
```diff
@@ -116,52 +130,291 @@ export class ViabilityService {
...
+    officeLevel: string,
+    officeType: string,
+    electionPartisanship: boolean,
+    seats: number,
+    incumbent: boolean,
+    opponents: number,
+    openSeat: boolean,
+  ): { probExponent: number; probOfWin: number; rating: number } {
+    // Implements a logistic regression term by term with co-efficients hard-coded
+    let probExponent = -1.70059412
```
tomer-tgp, line 271: it will be nice to have all of these random numbers in an object or enum so we can be more descriptive and understand what they mean.

PR #145, opened 2025-03-10
WEB-3742: Update Pro Splash Page & Add Session Based Dialogs

oldfashionedrobot:
- adds basic session tracking for users (just counts sessions)
- uses  30 min as the cooldown between a session atm (that is what fullstory uses)


src/users/services/sessions.service.ts
```diff
@@ -0,0 +1,37 @@
...
+    private readonly users: UsersService,
+  ) {}
+
+  async trackSession(user: User) {
+    const currentTime = Date.now()
+
+    // Get current metadata
+    const currentMetaData = user.metaData || {}
+    const lastVisited = currentMetaData.lastVisited || 0
+    const sessionCount = currentMetaData.sessionCount || 0
```
tomer-tgp, line 19: can we add sessionCount to 
https://github.com/thegoodparty/gp-api/blob/develop/prisma/schema/user.jsonTypes.ts#L12

tomer-tgp (changes requested):

tomer-tgp (approved):

RavenHursT (approved):

PR #148, opened 2025-03-10
fix: register-429

tomer-tgp:


oldfashionedrobot (approved):

src/campaigns/services/campaigns.service.ts
```diff
@@ -248,12 +248,8 @@ export class CampaignsService extends createPrismaBase(MODELS.Campaign) {
     )
 
     if (!campaign) {
-      // every user should have a campaign
-      campaign = await this.createForUser(user)
       return {
-        status: CampaignStatus.onboarding,
-        slug: campaign.slug,
-        step: 1,
+        status: false,
```
RavenHursT, line 252: We really should just `throw new NotFoundException()` here, but approving for production issue

PR #155, opened 2025-03-11
WEB-3846 ecanvasser-sync

tomer-tgp:
Logic:
sync without force will only sync if the lastSync was 30 minutes ago or more.
Sync with force=true will sync ecanvasser 

this allows us to sync the data when the user with ecanvasser visits the dashboard so they see fresh data and also allow the user to force sync their data.


RavenHursT (approved):
Approving.. but I'd really love to see is move away from doing these kinds of semaphore tasks within the confines of an http request. Triggering with a request is fine, but that should just push an appropriate task into a queue, that should then be picked up in the background by a consumer.

PR #174, opened 2025-03-16
Web 3828 ecanvasser surveys

tomer-tgp:
https://github.com/user-attachments/assets/4e6b905a-af6d-4a1c-afd6-bbd88d89956b


src/ecanvasser/services/ecanvasser.service.ts
```diff
@@ -222,16 +234,37 @@ export class EcanvasserService extends createPrismaBase(MODELS.Ecanvasser) {
...
       return response.data as ApiResponse<T>
     } catch (error) {
-      this.logger.error(`Failed to fetch from ${endpoint}`, error)
+      this.logger.error(
+        `Failed to ${options.method || 'GET'} ${endpoint}`,
+        error,
+      )
       throw error
     }
   }
```
RavenHursT, line 270: It looks like we've got two "sources of truth" here for ecanvasser data, right?

On one hand, this service is implementing our `createPrismaBase` pattern that has a series of DB tables that are holding fields that we "care" about.

But on the other, we're sometimes going to this `fetchFromApi` method to interact w/ the _actual_ Ecanvasser API data.

This seems to be breaking the SoC Principle: https://www.geeksforgeeks.org/separation-of-concerns-soc/

It probably would behoove us to rename this module/service to something like `CampaignMetrics` or `CampaignOperations`, and then have a seperate `Ecanvasser` service that handling the actual integration w/ the Ecanvasser API.

We're GoodParty. We don't own/aren't Ecanvasser. We shouldn't be naming/mixing our business logic after/with third party vendors. 

(ie. see the `Content` and `Contentful` modules for an example of this pattern)

Doing this will help to reduce confusion, protect our code/DB from vendor lock-in should we choose to interact w/, or augment this data from a different vendor, as well as help simplify refactors in the future around this feature. This also encapsulates and normalizes the HTTP-specific language here, so that instead of using HTTP method and URL path shuffling to interact w/ the Ecanvasser API, we instead would be using simple CRUD verbs from that service.

src/ecanvasser/services/ecanvasser.service.ts
```diff
@@ -453,4 +486,252 @@ export class EcanvasserService extends createPrismaBase(MODELS.Ecanvasser) {
...
+          method: 'POST',
+          data: payload,
+        },
+      )
+
+      return response.data
+    } catch (error) {
+      this.logger.error('Failed to create survey question', error)
+      throw error
+    }
```
RavenHursT, line 574: All of these catches in these methods talking to the Ecanvasser API should be `throw`ing `BadGatewayException`, not just whatever is being coughed up from this nebulous `fetchFromApi` method.

src/ecanvasser/services/survey.service.ts
```diff
@@ -0,0 +1,279 @@
...
+import { UpdateSurveyQuestionDto } from '../dto/update-survey-question.dto'
+import { UpdateSurveyDto } from '../dto/update-survey.dto'
+
+@Injectable()
+export class SurveyService {
+  private readonly logger = new Logger(SurveyService.name)
+  private readonly apiUtil: EcanvasserApiUtil
+
+  constructor(
+    private readonly httpService: HttpService,
```
RavenHursT, line 21: This isn't used. Delete.

src/ecanvasser/services/ecanvasser.service.ts
RavenHursT: Wait.. isn't this, at least, in partial, basically a duplicate of the functionality that's in `ecanvasser.util.ts`? 🤔

src/ecanvasser/util/ecanvasser.util.ts
```diff
@@ -0,0 +1,131 @@
+import { HttpService } from '@nestjs/axios'
+import { Logger } from '@nestjs/common'
+import { lastValueFrom } from 'rxjs'
+import { ApiResponse, PaginationParams } from '../ecanvasser.types'
+
+const DEFAULT_PAGE_SIZE = 1000
+
+export class EcanvasserApiUtil {
```
RavenHursT, line 8: Yeah.. this definitely needs to not just be a "util" class.. 

This needs to be a`EvanvasserService` within a separate `EcanvasserModule`

src/ecanvasser/services/survey.service.ts
```diff
@@ -0,0 +1,279 @@
...
+    private readonly ecanvasserService: EcanvasserService,
+  ) {
+    this.apiUtil = new EcanvasserApiUtil(httpService)
```
RavenHursT, line 24: This is an antipattern in Nest or DI pattern systems in general.

If `EcanvasserService` was an actual service that interacts/implements the integration w/ the Ecanvasser API, then Nest would be providing a _singleton_ in memory for this service to then consume and interact with. No need to create and manage a separate instance here as a private member on _this_ service.

src/ecanvasser/services/survey.service.ts
```diff
@@ -0,0 +1,279 @@
...
+        {
+          method: 'POST',
+          data: payload,
+        },
+      )
+
+      return response.data
+    } catch (error) {
+      this.logger.error('Failed to create survey', error)
+      throw error
```
RavenHursT, line 54: These should also all be `BadGatewayExceptions` thrown in this service as well

RavenHursT (changes requested):

src/ecanvasser/services/ecanvasser.service.ts
```diff
@@ -197,9 +204,14 @@ export class EcanvasserService extends createPrismaBase(MODELS.Ecanvasser) {
   private async fetchFromApi<T>(
     endpoint: string,
     apiKey: string,
-    params: PaginationParams = {},
+    options: {
+      method?: 'GET' | 'POST' | 'PUT' | 'DELETE'
```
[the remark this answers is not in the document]
RavenHursT: Oh.. just had to just set the `Methods` type:
image

RavenHursT (approved):

PR #186, opened 2025-03-22
[WEB-3822] - simplify blog articles/fix

RavenHursT:
https://goodparty.atlassian.net/browse/WEB-3822

UI PR: https://github.com/thegoodparty/gp-webapp/pull/562

Postman Run: https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/34024107-bdc8cec4-9807-4e64-a2a6-5a0accd82bad


src/content/util/mapBlogArticlesToSections.util.ts
```diff
@@ -0,0 +1,17 @@
...
+    articles.reduce((acc, curr) => {
+      const currentSectionSlug = curr.section.fields.slug
+      return acc.set(currentSectionSlug, [
+        ...(acc.has(currentSectionSlug)
+          ? [...(acc.get(currentSectionSlug) || []), curr].slice(0, limit)
+          : [curr]),
+      ])
+    }, new Map<string, BlogArticleMeta[]>()),
+  )
+}
```
tomer-tgp, line 17: Can we make this more readable?
Something like this?

```
export const mapBlogArticlesToSections = (
  articles: BlogArticleMeta[],
  limit?: number,
) => {
  const sectionMap = new Map<string, BlogArticleMeta[]>();
  
  // Group articles by their section
  for (const article of articles) {
    const sectionSlug = article.section.fields.slug;
    
    // Get existing articles for this section or create empty array
    const existingArticles = sectionMap.get(sectionSlug) || [];
    
    const updatedArticles = [...existingArticles, article];
    
    const limitedArticles = limit ? updatedArticles.slice(0, limit) : updatedArticles;
    
    sectionMap.set(sectionSlug, limitedArticles);
  }
  
  return Object.fromEntries(sectionMap);
}

```

tomer-tgp (changes requested):

src/content/content.controller.ts
```diff
@@ -123,6 +115,36 @@ export class ContentController {
...
+        publishDate: 'desc',
+      },
+      ...(limit ? { take: limit } : {}),
+    })
+  }
+
+  @Get(['blog-articles/by-section/:sectionSlug', 'blog-articles/by-section'])
+  async listBlogArticleSummariesBySection(
+    @Param('sectionSlug') sectionSlug?: string,
+    @Query('limit') limit?: number, // Limit articles per section
```
oldfashionedrobot, line 133: I believe this might need a ParseIntPipe here too

oldfashionedrobot (changes requested):

src/content/util/mapBlogArticlesToSections.util.ts
RavenHursT: I don't think introducing and additional scope and more variables with a `for-loop` approach is warranted here, so I kept the `.reduce` approach.

I was however able to break up the return value and was actually able to reduce the cyclical complexity since there's no need to check for `limit` since `.slice(0, undefined)` will give the entire array anyway, and spreading `[...[]]` results in `[]` anyway, so no need to check for existing array, just append to the end, even if it's empty ("point-free" programming FTW!)

RavenHursT: https://github.com/thegoodparty/gp-api/pull/186/commits/5d33ef3aba928b7910505a0bd85c91c9074e1ca7

src/content/content.controller.ts
RavenHursT:
```typescript
console.log(`typeof limit =>`, typeof limit)
```
> `typeof limit => string`

Weird.. I wonder why TS didn't catch that when I've very clearly declared that the `limit` query param should be `number`? 🤔

oldfashionedrobot (approved):

tomer-tgp (approved):

PR #187, opened 2025-03-24
Web 3924 texting automation

tomer-tgp:
Front end PR: https://github.com/thegoodparty/gp-webapp/pull/563


oldfashionedrobot (changes requested):

src/textCampaign/schemas/textCampaign.schema.ts
```diff
@@ -0,0 +1,12 @@
+import { z } from 'zod'
+
+export enum TextCampaignStatus {
+  PENDING = 'pending',
+  APPROVED = 'approved',
+  DENIED = 'denied',
+  PAID = 'paid',
+  IN_PROGRESS = 'in_progress',
+  COMPLETED = 'completed',
+}
```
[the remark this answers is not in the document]
tomer-tgp: oh, I wasn't even using this file. Fixed.

src/textCampaign/services/textCampaign.service.ts
```diff
@@ -0,0 +1,243 @@
...
+import { TextCampaignSummary } from '../textCampaign.types'
+import { TextCampaignStatus } from '@prisma/client'
+
+@Injectable()
+export class TextCampaignService {
+  public readonly logger = new Logger(TextCampaignService.name)
+
+  constructor(
+    @Inject(forwardRef(() => CampaignsService))
+    private readonly campaignsService: CampaignsService,
```
RavenHursT, line 21: This really feels like a code-smell here.

We're injecting the entire `CampaignService` here, cyclically even, just to do what amounts to determining if a given `campaignId` actually exists, right (down in `createProject`, and _maybe_ `uploadContactsAndCreateProject` if that dead code isn't removed for the time being).

Seems like there could be a better ACL-type solution here?

Maybe we need to create a custom NestJS "guard" to make this determination up in the controller, as opposed to down here in the service? 🤔

src/textCampaign/services/textCampaign.service.ts
```diff
@@ -0,0 +1,243 @@
...
+
+    // Step 2: Create a project using the group ID
+    const createProjectDto: CreateProjectSchema = {
+      ...projectDetails,
+      groupId,
+    }
+
+    // Step 3: Create the project with the group ID
+    return this.createProject(campaignId, createProjectDto)
+  }
```
RavenHursT, line 123: Also.. when if and when we _do_ implement this...

This kind of orchestration between operations (upload, and then create) should be handled up in the controller layer, not down here in the service layer.

Have a `createContactGroup` method that takes the CSV, sends it to the import method in the `RumbleUpService` and returns the newly created `groupId`

Then, in the controller endpoint method, just call `createContactGroup`, and pass that `groupId` along to `createProject`

src/textCampaign/textCampaign.types.ts
```diff
@@ -0,0 +1,54 @@
...
+export interface ApiRumbleUpResponse {
+  success: boolean
+  data?: any
+  message?: string
+  error?: string
+}
+
+export interface CreateProjectResponse {
+  success: boolean
+  data?: any
```
RavenHursT, line 30: Please type these `any`s

src/voters/services/voterOutreach.service.ts
```diff
@@ -123,6 +125,28 @@ export class VoterOutreachService {
...
+          campaign.id,
+          `SMS Campaign ${new Date(date).toLocaleDateString()}`,
+          message,
+          audience,
+          messagingScript,
+          new Date(date),
+          imageUrl,
+        )
+
+        this.logger.log(`Created TextCampaign for campaign ${campaign.id}`)
```
RavenHursT, line 141: logger.debug

src/voters/services/voterOutreach.service.ts
```diff
@@ -123,6 +125,28 @@ export class VoterOutreachService {
...
+      } catch (error: any) {
+        this.logger.error(
+          `Failed to create TextCampaign: ${error.message}`,
+          error.stack,
+        )
+      }
+    }
```
RavenHursT, line 148: Not sure this `try-catch` is needed. This error will likely just be redundant since the service we're calling will be throwing it's own exceptions, no?

Also.. we never re-throw here, so any exception thrown out of the `createTextCampaign` will just get swallowed here.

Prolly best to just let the exception bubble up..

src/voters/voterFile/schemas/ScheduleOutreachCampaign.schema.ts
```diff
@@ -12,15 +12,14 @@ export class ScheduleOutreachCampaignSchema extends createZodDto(
         .object(
           CUSTOM_FILTERS.reduce(
             (acc, filterKey) => {
-              acc[filterKey] = z.boolean().optional()
+              acc[filterKey] =
+                filterKey === 'audience_request'
```
RavenHursT, line 16: `enum`?

RavenHursT: Oh.. it's already right here.. use this: https://github.com/thegoodparty/gp-api/pull/187/files#diff-6e50362ee30dfcdd287568bf0dbe361431bc6e8a2afaa96378d9889e510d5d73R50

RavenHursT (changes requested):

src/voters/voterFile/schemas/ScheduleOutreachCampaign.schema.ts
tomer-tgp: That is an array, so it get's a little dirty trying to reuse it.

src/textCampaign/services/rumbleUp.service.ts
```diff
@@ -0,0 +1,65 @@
...
+      `Failed to make request to RumbleUp API`,
+      error.response?.data || error.message,
+    )
+
+    if (error.response?.status === 401) {
+      throw new BadGatewayException(
+        'Unauthorized: Invalid RumbleUp credentials',
+      )
+    }
+    if (error.response?.status === 429) {
```
RavenHursT, line 57: `HttpStatus` constants please.

src/textCampaign/services/textCampaign.service.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    // Call RumbleUp API to create the project
+    const response =
+      await this.rumbleUpService.createProject(rumbleUpProjectData)
+
+    if (!response.success) {
+      throw new Error(
+        `Failed to create project in RumbleUp: ${response.error || response.message}`,
+      )
+    }
```
RavenHursT, line 40: This should throw a `BadGatewayException` no? 🤔

src/textCampaign/services/textCampaign.service.ts
```diff
@@ -0,0 +1,126 @@
...
+import { RumbleUpService } from './rumbleUp.service'
+import { TextCampaignStatus } from '@prisma/client'
+
+@Injectable()
+export class TextCampaignService {
+  public readonly logger = new Logger(TextCampaignService.name)
+
+  constructor(
+    private readonly rumbleUpService: RumbleUpService,
+    private readonly prisma: PrismaService,
```
RavenHursT, line 13: If you have this class use `createPrismaBase` then you don't have to declare a logger and you don't have to inject the `PrismaService`. It's all done for you.

https://github.com/thegoodparty/gp-api/blob/8f4d0a99264f0705aa0c616050b97744d95192c8/src/campaigns/services/campaigns.service.ts#L29

src/textCampaign/services/textCampaign.service.ts
```diff
@@ -0,0 +1,126 @@
...
+          gender_unknown: audience.gender_unknown,
+          audience_request: audience.audience_request,
+        }),
+        script,
+        date,
+        imageUrl,
+      },
+    })
+  }
+}
```
RavenHursT, line 126: If you use `createPrismaBase`, then you don't have to define this method at all. You can just call `this.textCampaignService.create()` from the `VoterOutreachService` directly.

RavenHursT (changes requested):

src/voters/services/voterOutreach.service.ts
```diff
@@ -123,6 +125,21 @@ export class VoterOutreachService {
       audienceRequest: audience['audience_request'],
     })
 
+    // If type is SMS, create a TextCampaign
+    if (type === 'sms') {
```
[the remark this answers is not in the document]
tomer-tgp: This is a very different service. These are not our regular transactional SMS but messages to voters paid by the candidate.

src/voters/services/voterOutreach.service.ts
```diff
@@ -123,6 +125,41 @@ export class VoterOutreachService {
...
+      await this.textCampaignService.model.create({
```
RavenHursT, line 130: You shouldn't have to reference `.model` here.. 

```suggestion
      await this.textCampaignService.create({
```

Should work.. Not blocking tho

RavenHursT (approved):

PR #215, opened 2025-04-02
HubSpot fix: use correct election date validation

oldfashionedrobot:


src/campaigns/services/crmCampaigns.service.ts
```diff
@@ -128,6 +128,8 @@ export class CrmCampaignsService {
   }
 
   private async createCompany(companyObj: CRMCompanyProperties) {
+    if (!IS_PROD) return
```
RavenHursT, line 131: We're hoping to have a more comprehensive solution to this feature switching:

https://goodparty.atlassian.net/browse/WEB-3952

RavenHursT (approved):

PR #245, opened 2025-04-15
WEB-3995: update ai content template data / api response

oldfashionedrobot:
- adds ai content tempalte ids from contentful
- removes skipVoterCount, (can just check for education task type instead)
- add `taskOnly` field and `id` to ai content template response

### DIFF
--- a/src/campaigns/tasks/fixtures/tasks-week-1.ts
+++ b/src/campaigns/tasks/fixtures/tasks-week-1.ts
@@ -10,6 +10,7 @@ const tasksWeek1: CampaignTask[] = [
     flowType: CampaignTaskType.text,
     deadline: 3,
     proRequired: true,
+    defaultAiTemplateId: '5b6W9pYlX796TBI2HV7HlQ',
   },
   {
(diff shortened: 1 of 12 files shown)


src/campaigns/tasks/fixtures/tasks-week-1.ts
```diff
@@ -47,6 +51,7 @@ const tasksWeek1: CampaignTask[] = [
     cta: 'Write Post',
     week: 1,
     flowType: CampaignTaskType.socialMedia,
+    defaultAiTemplateId: 'GpWsRql46Nif2wYroxj81',
```
RavenHursT, line 54: This is gonna be very brittle given that Contentful users can delete these on a whim 😕

RavenHursT (approved):

src/campaigns/tasks/fixtures/tasks-week-1.ts
oldfashionedrobot: I think we can do a better solution in a followup iteration. Justin made those templates, so I don't think anyone will delete them randomly

PR #250, opened 2025-04-17
WEB-4032: added backend min date validation for scheduled campaigns

oldfashionedrobot:


src/voters/voterFile/schemas/ScheduleOutreachCampaign.schema.ts
```diff
@@ -25,7 +26,22 @@ export class ScheduleOutreachCampaignSchema extends createZodDto(
...
+          const minDate = startOfDay(addDays(new Date(), 3))
+          return (
+            isAfter(selectedDate, minDate) ||
+            selectedDate.getTime() === minDate.getTime()
+          )
+        },
+        {
+          message: 'Date must be at least 72 hours from now',
+        },
+      ),
```
RavenHursT, line 44: I think we may need to base this off of the current "now" date from the client?

Our API is running in Oregon, there could be some discrepancy here if someone is making a request from another time zone that is closer to the day boundary like the East Coast or Hawaii, right? 🤔

RavenHursT (changes requested):

src/voters/voterFile/schemas/ScheduleOutreachCampaign.schema.ts
oldfashionedrobot: the date param comes in as just `yyyy-mm-dd` so im just doing it by calendar days to avoid time issues.

RavenHursT (approved):

PR #263, opened 2025-04-24
WEB-4090: validate campaign details json fields

oldfashionedrobot:
- I only added the validation for the details fields for now, in case this causes any other issues. We can add validation like this to all our jsonb fields if it works well

### DIFF
--- a/src/campaigns/schemas/updateCampaign.schema.ts
+++ b/src/campaigns/schemas/updateCampaign.schema.ts
@@ -71,7 +71,7 @@ const CampaignDetailsSchema = z
   .strict()
   .partial()
 
-// TODO: make schemas for the actual JSON content
+// TODO: make schemas data, pathToVictory, aiContent
 export class UpdateCampaignSchema extends createZodDto(
   z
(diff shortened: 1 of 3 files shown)


src/campaigns/schemas/updateCampaign.schema.ts
```diff
@@ -1,13 +1,83 @@
...
+    officeTermLength: z.string(),
+    partisanType: z.string(),
+    priorElectionDates: z.array(z.string()),
+    positionId: z.string(),
+    tier: z.string(),
+  })
+  .strict()
+  .partial()
 
 // TODO: make schemas for the actual JSON content
```
RavenHursT, line 74: We can delete this comment now, correct?

RavenHursT (approved):

src/campaigns/schemas/updateCampaign.schema.ts
oldfashionedrobot: oh, I'll update to be for the other three json fields there.

PR #288, opened 2025-05-08
WEB-4084 - Setup segment

oakinh:
image


src/fullStory/fullStory.module.ts
```diff
@@ -2,11 +2,12 @@ import { Module } from '@nestjs/common'
 import { FullStoryService } from './fullStory.service'
 import { HttpModule } from '@nestjs/axios'
 import { FullStoryController } from './fullStory.controller'
+import { SegmentModule } from 'src/segment/segment.module'
```
tomer-tgp, line 5: These should not be combined.
We will deprecate fullStory very soon.

.env.example
```diff
@@ -63,6 +63,9 @@ SLACK_BOT_POLITICS_CHANNEL_TOKEN="SAMPLE_POLITICS_CHANNEL_TOKEN"
 SLACK_BOT_FEEDBACK_CHANNEL_ID="SAMPLE_FEEDBACK_CHANNEL_ID"
 SLACK_BOT_FEEDBACK_CHANNEL_TOKEN="SAMPLE_FEEDBACK_CHANNEL_TOKEN"
 
+SEGMENT_WRITE_KEY_PROD=key
+SEGMENT_WRITE_KEY_DEV=key
```
tomer-tgp, line 67: you probably just need `SEGMENT_WRITE_KEY` instead of these two

tomer-tgp (changes requested):

src/fullStory/fullStory.module.ts
oakinh: I did this so we don't have to replicate all of the logic in the segment module. When we deprecate fullstory I'll just have the Segment Module become what is currently the FullStory Module. Cuts down the work and opportunity for mistakes that way, no?

.env.example
oakinh: Yep that's what I did, forgot to update the .env.example, fixing

src/fullStory/fullStory.module.ts
tomer-tgp: Maybe rename it to analytics module then?

oakinh: Renamed

tomer-tgp (approved):

PR #319, opened 2025-06-20
DT 73 - Segment Analytics Phase 2

oakinh:


src/payments/services/paymentEventsService.ts
```diff
@@ -241,6 +244,24 @@ export class PaymentEventsService {
     })
     await this.campaignsService.setIsPro(campaignId)
 
+    // For Segment
+    const sub = session.subscription as Stripe.Subscription
```
RavenHursT, line 248:
```suggestion
    const subscription = session.subscription as Stripe.Subscription
```

Just makes it explicit what we're referring to

src/payments/services/paymentEventsService.ts
```diff
@@ -241,6 +244,24 @@ export class PaymentEventsService {
...
+    const pm = intent.payment_method as Stripe.PaymentMethod
+
+    const paymentMethod =
+      pm.type === 'card' ? (pm.card?.wallet?.type ?? 'credit card') : pm.type
+
+    this.segment.trackEvent(user.id, EVENTS.Account.ProSubscriptionConfirmed, {
+      price,
+      paymentMethod,
+      renewalDate: new Date(sub.current_period_end * 1000).toISOString(),
+    })
```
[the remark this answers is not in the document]
RavenHursT: They can also _regain_ their `isPro` flag through reactivation of their subscription through the Stripe webhook: 

https://github.com/thegoodparty/gp-api/blob/7e291f4b516dd35e0ecd9fd62dfb3e8b2f9ac8b1/src/payments/services/paymentEventsService.ts#L96-L140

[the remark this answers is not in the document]
oakinh: This would probably be a different event, I'll ask Bryan about this and if he wants it done, whether the same or a different event, I'll have him add it to the Phase 3 ticket so this one isn't blocked

src/analytics/analytics.service.ts
```diff
@@ -231,7 +231,6 @@ export class AnalyticsService {
     user: User,
     properties: TrackingProperties,
   ) {
-    this.segment.identify(user.id, properties)
```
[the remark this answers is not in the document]
oakinh: Required a lot of forward refs, but it's done

src/analytics/analytics.module.ts
```diff
@@ -1,13 +1,23 @@
...
 @Module({
   providers: [AnalyticsService],
   exports: [AnalyticsService],
-  imports: [HttpModule, SegmentModule],
+  imports: [
+    HttpModule,
+    SegmentModule,
+    forwardRef(() => SharedModule),
+    forwardRef(() => CampaignsModule),
+    forwardRef(() => UsersModule),
```
RavenHursT, line 19: None of these are needed.

Please go through the rest of these changes and try to avoid these. Rearrange how modules, controllers, and services are being imported/exported if need be.

Unfortunately the `forwardRef` tool ends up being a pretty big temptation to use as an "escape hatch", when it's typically avoidable.. Something I've had to think on and learn, myself in the past few months since using NestJS..

RavenHursT (changes requested):
Let me know when you want me to take another look 👍

src/analytics/analytics.module.ts
oakinh: Ok I managed to remove some of the forwardRefs, but not all. The ones remaining break if not present.

If these are a problem, here's a few solutions I'm thinking of:
1) Create new "provider" services in the UsersModule and CampaignsModule that's just another layer of abstraction on top of functions like findUser, patchUserMetaData, findMany, findByUserId, etc
2) Separate the fullstory and segment logic like it was before, since Segment doesn't have any other service or module dependencies
3) Add a 3rd module layer on top of Analytics so it's MetaAnalytics -> Analytics -> Segment
4) Keep Analytics as a module just for fullstory, and create a separate module called AnalyticsRouting that sits on top of just the Segment Module

All of these except maybe #2 sound like bad options to me, mostly because we're supposed to rip out fullstory soon, at which time Analytics won't have any circular dependencies with just Segment logic.

Thoughts?

RavenHursT: Nah.. at least they're separated out now and that should help us out a bit when we rip out FullStory to refactor.

And it looks like once that's done, most of the need for `CampaignsService` and `UsersService` in the `AnalyticsService` will be moot, so we'll be able to remove them then.

RavenHursT (approved):
Consider merging this: https://github.com/thegoodparty/gp-api/pull/325

before merging this PR.

PR #320, opened 2025-06-20
WEB-4251 

Collin-GP:
add a migration script to delete scheduled messages that we don't want to send anymore.


RavenHursT (approved):

Collin-GP:
I think either of these methods works, since this one is such a simple case, 
but i think not running scripts directly against the db is good practice as well, 
at least this way we all see the script being run (accountability and transparency)

PR #328, opened 2025-07-01
WEB-4300: [Usersnap] Never stops loading

oldfashionedrobot:
add timeout for races searching, pull state from zip code as fallback filter
See: https://goodparty.atlassian.net/browse/WEB-4300?focusedCommentId=19882

### DIFF
--- a/src/elections/services/races.service.ts
+++ b/src/elections/services/races.service.ts
@@ -79,7 +79,7 @@ export class RacesService {
         }
 
         iterationCount++
-        this.logger.log(
+        this.logger.debug(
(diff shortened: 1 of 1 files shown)


src/elections/services/ballotReady.service.ts
```diff
@@ -15,6 +15,7 @@ import {
 } from '../types/ballotReady.types'
 import { Headers, MimeTypes } from 'http-constants-ts'
 import { ELECTION_LEVELS } from '../../shared/constants/governmentLevels'
+import zipcodes from 'zipcodes'
```
RavenHursT, line 18: 👏 ❤️

PR #340, opened 2025-07-08
feat: SRV-17 adding a uuid to community issues

tomer-tgp:
This is for the detailed view of the issue. The UUID needs to be in the browser URL, and I didn't want to use the DB ID.

### DIFF
--- a/prisma/migrations/20250708021856_add_uuid_to_community_issue/migration.sql
+++ b/prisma/migrations/20250708021856_add_uuid_to_community_issue/migration.sql
@@ -1,0 +1,15 @@
+/*
+  Warnings:
+
+  - A unique constraint covering the columns `[uuid]` on the table `community_issue` will be added. If there are existing duplicate values, this will fail.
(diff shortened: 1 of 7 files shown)


israelvicars (commented):
@tomer-tgp Looks like the controller and service changes will still work without the redundant index.

In prisma/migrations/20250708021856_add_uuid_to_community_issue/migration.sql, lines 13-15.

In prisma/schema/communityIssue.prisma, line 32.

tomer-tgp:
@israelvicars it will work without the index, but will be slower, wouldn't it? We query with the UUID, so it should be indexed. Right?

oldfashionedrobot (approved):

RavenHursT:
Yeah.. def don't need the extra `@@index`

RavenHursT (changes requested):
Seems like the primary key should just be a UUID.. why add an additional unique key identifier when the UUID could just be the primary key? 🤔

RavenHursT (approved):

PR #359, opened 2025-07-20
Web 4233 domain purchase

tomer-tgp:
https://github.com/thegoodparty/gp-webapp/pull/729/

### DIFF
--- a/src/websites/services/domains.service.ts
+++ b/src/websites/services/domains.service.ts
@@ -29,10 +29,17 @@ export enum DomainOperationType {
   REGISTER_DOMAIN = 'RegisterDomain',
 }
 
+export interface DomainOperationDetail {
+  operationId: string | null
+  status: DomainOperationStatus
+  type: DomainOperationType
+  submittedDate: Date
+}
+
 export interface DomainStatusResponse {
   message: DomainOperationStatus
   paymentStatus: PaymentStatus | null
-  operationDetail?: any
+  operationDetail?: DomainOperationDetail
 }
 
 @Injectable()
@@ -96,7 +103,7 @@ export class DomainsService
       )
     }
 
-    const { paymentIntent, user } =
+    const { paymentIntent: _paymentIntent, user } =
       await this.payments.getValidatedPaymentUser(paymentIntentId)
 
     const validWebsiteId = this.convertWebsiteIdToNumber(websiteId)
@@ -106,11 +113,17 @@ export class DomainsService
       select: { content: true },
     })
 
+    const searchResult = await this.searchForDomain(domainName!)
+
+    if (!searchResult.price) {
+      throw new BadRequestException('Could not get price for domain')
+    }
+
     const domain = await this.model.create({
       data: {
         websiteId: validWebsiteId,
         name: domainName!,
-        price: paymentIntent.amount / 100,
+        price: searchResult.price,
         paymentId: paymentIntentId,
         status: DomainStatus.pending,
       },
@@ -153,20 +166,30 @@ export class DomainsService
   }
 
   private buildContactInfo(
-    user: any,
-    websiteContent: any,
+    user: User,
+    websiteContent: PrismaJson.WebsiteContent | null,
   ): RegisterDomainSchema {
-    const address = websiteContent?.contact?.address
+    const addressPlace = websiteContent?.contact?.addressPlace
     return {
-      firstName: user.firstName,
-      lastName: user.lastName,
-      email: user.email,
+      firstName: user.firstName || GP_DOMAIN_CONTACT.firstName,
+      lastName: user.lastName || GP_DOMAIN_CONTACT.lastName,
+      email: user.email || GP_DOMAIN_CONTACT.email,
       phoneNumber: user.phone || GP_DOMAIN_CONTACT.phoneNumber,
-      addressLine1: address?.addressLine1 || GP_DOMAIN_CONTACT.addressLine1,
-      addressLine2: address?.addressLine2 || GP_DOMAIN_CONTACT.addressLine2,
-      city: address?.city || GP_DOMAIN_CONTACT.city,
-      state: address?.state || GP_DOMAIN_CONTACT.state,
-      zipCode: address?.zipCode || GP_DOMAIN_CONTACT.zipCode,
+      addressLine1:
+        addressPlace?.formatted_address || GP_DOMAIN_CONTACT.addressLine1,
+      addressLine2: GP_DOMAIN_CONTACT.addressLine2,
+      city:
+        addressPlace?.address_components?.find((c) =>
+          c.types.includes('locality'),
+        )?.long_name || GP_DOMAIN_CONTACT.city,
+      state:
+        addressPlace?.address_components?.find((c) =>
+          c.types.includes('administrative_area_level_1'),
+        )?.short_name || GP_DOMAIN_CONTACT.state,
+      zipCode:
+        addressPlace?.address_components?.find((c) =>
+          c.types.includes('postal_code'),
+        )?.long_name || GP_DOMAIN_CONTACT.zipCode,
     }
   }
 
@@ -252,7 +275,7 @@ export class DomainsService
 
     const paymentIntent = await this.payments.createPayment(user, {
       type: PaymentType.DOMAIN_REGISTRATION,
-      amount: domain.price.toNumber() * 100, // convert to cents
+      amount: domain.price.toNumber() * 100,
       domainName,
       domainId: domain.id,
     })


tomer-tgp:
bugbot run

tomer-tgp:
bugbot run

tomer-tgp:
bugbot run

src/shared/util/numbers.util.ts
```diff
@@ -6,3 +6,56 @@ export const getRandomInt = (min: number, max: number = LARGEST_SAFE_INTEGER) =>
...
+      throw new Error('Non-US phone number and no fallback available')
+    }
+  } else {
+    // Invalid length, use fallback
+    if (fallbackNumber) {
+      return fallbackNumber
+    }
+    throw new Error('Invalid phone number length and no fallback available')
+  }
+}
```
RavenHursT, line 61: Seems like this util should just replace the other `formatPhoneNumber` we already have in the `AwsModule`, that's specific to the expected format for AWS services, rather than add one here in a `numbers.util.ts` that has two args that are both `string` and also the method returns `string`, not `number` in anycase.

https://github.com/thegoodparty/gp-api/blob/develop/src/aws/util/awsRoute53.util.ts#L2-L14

src/stripe/services/stripe.service.ts
```diff
@@ -36,6 +36,11 @@ export class StripeService {
     const userId = user.id
     const customerId = user.metaData?.customerId
 
+    // Filter out undefined values from metadata as Stripe doesn't accept them
+    const filteredMetadata = Object.fromEntries(
+      Object.entries(restMetadata).filter(([_, value]) => value !== undefined),
+    )
+
```
RavenHursT, line 43: If we end up w/ a consumer of this passing `undefined` values.. do we really want this error state to be swallowed?

Seems like we'd want Stripe to tell us if we're trying to send over these properties? 🤔

src/vercel/services/vercel.service.ts
```diff
@@ -9,6 +9,18 @@ if (!VERCEL_TOKEN || !VERCEL_PROJECT_ID) {
...
+  firstName: 'Victoria',
+  lastName: 'Mitchell',
+  email: 'accounts@goodparty.org',
+  phoneNumber: '+1.3126851162',
+  addressLine1: '916 Silver Spur Rd',
+  addressLine2: '',
+  city: 'Rolling Hills Estates',
+  state: 'CA',
+  zipCode: '90274',
+}
```
RavenHursT, line 22: Move to `vercel.consts.ts` that we've got in parent folder?

src/vercel/services/vercel.service.ts
```diff
@@ -23,9 +35,133 @@ export class VercelService {
...
+      const finalContact = {
+        firstName: contact.firstName || GP_DOMAIN_CONTACT.firstName,
+        lastName: contact.lastName || GP_DOMAIN_CONTACT.lastName,
+        email: contact.email || GP_DOMAIN_CONTACT.email,
+        phone: contact.phoneNumber || GP_DOMAIN_CONTACT.phoneNumber,
+        address1: contact.addressLine1 || GP_DOMAIN_CONTACT.addressLine1,
+        city: contact.city || GP_DOMAIN_CONTACT.city,
+        state: contact.state || GP_DOMAIN_CONTACT.state,
+        postalCode: contact.zipCode || GP_DOMAIN_CONTACT.zipCode,
+      }
```
RavenHursT, line 122: This seems duplicative.. Shouldn't this be the same logic as what's in `DomainPurchaseHandler::buildContactInfo`? 🤔 

Prolly apropos to my comment there.

src/websites/controllers/domains.controller.ts
```diff
@@ -55,44 +56,50 @@ export class DomainsController {
...
-  @UseCampaign()
-  async checkRegistrationStatus(@ReqCampaign() { id: campaignId }: Campaign) {
-    return this.domains.checkRegistrationStatus(campaignId)
+    return {
+      message,
+      paymentStatus,
+      operationDetail: {
+        operationId: domain.operationId,
+        status: message,
+        type: 'RegisterDomain',
```
RavenHursT, line 99: Should this be in an `enum` somewhere?

Not sure what this is used for since it's a static value that doesn't appear to change for this given endpoint 🤔

src/websites/controllers/domains.controller.ts
```diff
@@ -103,4 +110,37 @@ export class DomainsController {
...
+    const website = await this.websites.findUnique({
+      where: { campaignId },
+      select: {
+        id: true,
+        domain: { select: { status: true } },
+      },
+    })
+
+    if (!website) {
+      throw new BadRequestException('No website found for this campaign')
```
RavenHursT, line 127: Would this not be a `NotFoundException`? 🤔

src/websites/controllers/domains.controller.ts
```diff
@@ -103,4 +110,37 @@ export class DomainsController {
...
+    }
+
+    if (!website.domain) {
+      throw new BadRequestException('No domain found for this campaign')
```
RavenHursT, line 131: Same here

src/websites/schemas/UpdateWebsite.schema.ts
```diff
@@ -2,6 +2,55 @@ import { WebsiteStatus } from '@prisma/client'
...
+  place_id: z.string(),
+  plus_code: GooglePlusCodeSchema.optional(),
+  reference: z.string().optional(),
+  types: z.array(z.enum(GOOGLE_ADDRESS_PLACE_VALUES)),
+  url: z.string().optional(),
+  utc_offset: z.union([z.number(), z.string()]).optional(),
+  vicinity: z.string().optional(),
+  html_attributions: z.array(z.string()).optional(),
+  utc_offset_minutes: z.union([z.number(), z.string()]).optional(),
+})
```
RavenHursT, line 53: I'm wondering if these, along w/ the underlying GooglePlaces types in `website.jsonTypes` should just live in a `GoogleModule` to be importable in other modules... 🤔 

I struggled a bit w/ doing all this in my last PR since it doesn't appear to be anything out there for idiomatic server-side typing for GooglePlaces that _matches_ w/ Google's ui-side of things. So ultimate I decided not to do so.

So I'm glad you did this instead lol.. But now I'm thinking my code in the `TCRCompliance` stuff could benefit from it as well. But w/o having to import the entire `WebsiteModule`.

src/websites/services/domains.service.ts
```diff
@@ -5,22 +5,40 @@ import {
...
+// Enum for payment statuses (based on Stripe Payment Intent statuses)
+export enum PaymentStatus {
+  REQUIRES_PAYMENT_METHOD = 'requires_payment_method',
+  REQUIRES_CONFIRMATION = 'requires_confirmation',
+  REQUIRES_ACTION = 'requires_action',
+  PROCESSING = 'processing',
+  REQUIRES_CAPTURE = 'requires_capture',
+  CANCELED = 'canceled',
+  SUCCEEDED = 'succeeded',
+}
```
RavenHursT, line 25: Seems like something that should be down in `PaymentsModule`, not here.

src/websites/services/domains.service.ts
```diff
@@ -32,50 +50,59 @@ export class DomainsService extends createPrismaBase(MODELS.Domain) {
...
+    )
 
     return {
       domainName,
       availability: availabilityResp.Availability,
       prices: {
-        registration: searchedTldPrices?.registrationPrice,
-        renewal: searchedTldPrices?.renewalPrice,
+        registration: searchedDomainPrice,
+        renewal: searchedDomainPrice,
```
RavenHursT, line 105: Why not just return `price` property here w/ either `price` or `searchedDomainPrice` here? 

Why the two properties, when there's ever just one value, `price`, for a given request? 🤔

src/websites/services/domains.service.ts
```diff
@@ -104,9 +131,13 @@ export class DomainsService extends createPrismaBase(MODELS.Domain) {
       },
     })
 
+    if (!domain.price) {
+      throw new BadGatewayException('Domain price not available')
```
RavenHursT, line 135: 👍

src/websites/services/domains.service.ts
```diff
@@ -143,78 +174,145 @@ export class DomainsService extends createPrismaBase(MODELS.Domain) {
...
+        where: { id: domain.id },
+        data: { status: DomainStatus.inactive },
+      })
+
+      throw new BadRequestException(
+        `Failed to register domain with Vercel: ${
+          error instanceof Error ? error.message : 'Unknown error'
+        }`,
+      )
+    }
```
RavenHursT, line 226: BadGateway?

src/websites/services/domains.service.ts
```diff
@@ -143,78 +174,145 @@ export class DomainsService extends createPrismaBase(MODELS.Domain) {
...
+      }
+    } catch (error) {
+      this.logger.error('Error configuring domain:', error)
+      throw new BadRequestException(
+        `Failed to configure domain: ${
+          error instanceof Error ? error.message : 'Unknown error'
+        }`,
+      )
+    }
+  }
```
RavenHursT, line 257: This `try-catch` should only be around the call to vercel.. not _also_ the DB update (we already have the exception handling in place w/ Prisma), and it should be a `BadGateway`

Not sure if A.I. is spitting out all these `BadRequestException`s? 🤔 

Maybe good opportunity for a Cursor rule? 🤷

src/websites/services/domains.service.ts
```diff
@@ -143,78 +174,145 @@ export class DomainsService extends createPrismaBase(MODELS.Domain) {
...
+      const paymentIntent = await this.payments.retrievePayment(paymentId)
+      return paymentIntent.status as PaymentStatus
+    } catch (error) {
       this.logger.warn(
-        `Domain ${domain.name} added to Vercel but requires verification`,
-        vercelResponse.verification,
+        `Failed to retrieve payment status for ${paymentId}:`,
+        error,
       )
+      return null
```
RavenHursT, line 268: We want this to fail silently?

Shouldn't a call to this w/ a bad `paymentId` return a 404, or at least a BadGateway? 🤔

src/shared/util/numbers.util.ts
```diff
@@ -6,3 +6,56 @@ export const getRandomInt = (min: number, max: number = LARGEST_SAFE_INTEGER) =>
 
 export const getRandomPercentage = () =>
   faker.number.float({ min: 0, max: 100, fractionDigits: 2 })
+
+export const getRandomElementFromArray = (array: any[]) =>
+  array[getRandomInt(0, array.length - 1)]
```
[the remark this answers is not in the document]
tomer-tgp: I think that was a merge conflict error

src/payments/handlers/domain-purchase.handler.ts
```diff
@@ -0,0 +1,163 @@
...
+      )
+    }
+
+    const user = await this.usersService.findUser({
+      id: parseInt(userId),
+    })
+
+    if (!user) {
+      throw new BadRequestException('User not found for payment')
+    }
```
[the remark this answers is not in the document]
RavenHursT: Ah.. didn't realize that there are two different "metadata" in this method.. confusing.

But still.. if we have a website loaded from the DB based off the `websiteId`, we can derive the `userId` from the user associated w/ the website, no?

tomer-tgp:
bugbot run

src/payments/handlers/domain-purchase.handler.ts
```diff
@@ -0,0 +1,163 @@
...
+    if (paymentIntent.status !== 'succeeded') {
+      throw new BadRequestException(
+        `Payment not completed. Status: ${paymentIntent.status}`,
+      )
+    }
+
+    const userId = paymentIntent.metadata?.userId
+    if (!userId) {
+      throw new BadRequestException('No userId found in payment metadata')
+    }
```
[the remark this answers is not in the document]
RavenHursT: This got moved down to the `PaymentsService`, but it's still a `BadRequestException` when it should prolly be a `BadGatewayException`

https://github.com/thegoodparty/gp-api/pull/359/files#diff-9e84329aaf79ff08fd11b32fc7e9d405542aaa8ac0eaee2044b4dcb022273e9bR35-R37

src/websites/services/domains.service.ts
```diff
@@ -31,52 +48,177 @@ export class DomainsService extends createPrismaBase(MODELS.Domain) {
...
+      throw new BadRequestException('Domain name is required')
+    }
+
+    const searchResult = await this.searchForDomain(domainName)
+
+    if (!searchResult.price) {
+      throw new BadRequestException('Could not get price for domain')
+    }
+
+    return searchResult.price * 100
```
RavenHursT, line 78: This is still being multiplied in the service method?

#359 (comment)

src/websites/services/domains.service.ts
```diff
@@ -31,52 +48,177 @@ export class DomainsService extends createPrismaBase(MODELS.Domain) {
...
+    paymentIntentId: string,
+    metadata: PurchaseMetadata,
+  ): Promise<any> {
+    return this.handleDomainPostPurchase(paymentIntentId, metadata)
+  }
+
+  async handleDomainPostPurchase(
+    paymentIntentId: string,
+    metadata: any,
+  ): Promise<any> {
```
RavenHursT, line 91: We still have an anys here?

#359 (comment)

PR #379, opened 2025-07-30
feat: base purchase on stripe API key

tomer-tgp:


RavenHursT (approved):
Not blocking.. just seems like some overkill or odd logic, possibly left over from AI?

PR #382, opened 2025-08-02
chore: add houses count to hubspot

tomer-tgp:


israelvicars (approved):
Approved pending bugbot fix

PR #390, opened 2025-08-04
fix: update dockerfile

tomer-tgp:
# Fix ECS Deployment Timeout Issue

## Problem

Deployment was failing with ECS service timeout error:
```
Waiter ServicesStable failed: Max attempts exceeded
```

The ECS service was not reaching a stable state because health checks were failing.

## Root Cause

[PR #389](https://github.com/thegoodparty/gp-api/pull/389) restructured the Prisma schema to use **multi-file schemas** with the new `prismaSchemaFolder` preview feature:

- **Schema files moved**: From single `prisma/schema.prisma` to multiple files in `prisma/schema/` directory
- **Migrations relocated**: From `prisma/migrations/` to `prisma/schema/migrations/` 
- **New structure**: Uses `previewFeatures = ["prismaSchemaFolder"]` in `prisma/schema/schema.prisma`

However, the **Dockerfile was not updated** to handle this new schema structure, causing:
- Database migrations to fail during Docker build
- Prisma client generation to fail
- Application startup issues due to missing database schema
- Health check failures (health endpoint depends on database connectivity)

## Solution

Updated the Dockerfile to explicitly specify the new schema location:

**Before:**
```dockerfile
RUN echo "Cache bust: $CACHEBUST" && npm run migrate:deploy
RUN echo "Cache bust: $CACHEBUST" && npm run generate
```

**After:**
```dockerfile
RUN echo "Cache bust: $CACHEBUST" && npx prisma migrate deploy --schema=prisma/schema
RUN echo "Cache bust: $CACHEBUST" && npx prisma generate --schema=prisma/schema
```

## Why This Fixes the Issue

1. **Explicit schema path**: Prisma commands now correctly locate the schema files in `prisma/schema/`
2. **Proper migrations**: Database migrations run successfully during Docker build
3. **Correct client generation**: Prisma client generates with the right schema structure
4. **Healthy application startup**: App starts with proper database connectivity
5. **Passing health checks**: Health endpoint can successfully query the database
6. **ECS stability**: Service reaches stable state within timeout limits

## Testing

- ✅ Docker build completes successfully
- ✅ Database migrations apply correctly  
- ✅ Prisma client generates without errors
- ✅ Application starts and health checks pass
- ✅ ECS service reaches stable state

### DIFF
--- a/deploy/Dockerfile
+++ b/deploy/Dockerfile
@@ -31,10 +31,10 @@ ENV DATABASE_URL=$DATABASE_URL
 
 # Note: the CodeBuild project must be configured to be on the VPC.
 # Force cache bust for these commands by echoing CACHEBUST
-RUN echo "Cache bust: $CACHEBUST" && npm run migrate:deploy
+RUN echo "Cache bust: $CACHEBUST" && npx prisma migrate deploy --schema=prisma/schema
 
 # Force cache bust for Prisma client generation
-RUN echo "Cache bust: $CACHEBUST" && npm run generate
+RUN echo "Cache bust: $CACHEBUST" && npx prisma generate --schema=prisma/schema
 
 # Build the application (output in /dist)
 RUN echo "Cache bust: $CACHEBUST" && npm run build


Collin-GP (approved):
lgtm

RavenHursT:
Hold up... Why isn't that getting picked up in the package.json?

It's the same config.. I'm not seeing how this is gonna fix it...

Collin-GP:
> Hold up... Why isn't that getting picked up in the package.json?
> 
> https://github.com/thegoodparty/gp-api/blob/develop/package.json#L29
> 
> It's the same config.. I'm not seeing how this is gonna fix it...

i thought maybe in the docker env it might not resolve prisma.schema but calling it out explicitly may work, figured we could try and revert if it doesnt work?

PR #417, opened 2025-08-07
[WEB-4426] Order FAQ articles

israelvicars:
If there's an order property on an FAQ article, it can be used to order the FAQ article list. Article categories were already ordered, so this just adds that ordering convention to the articles.

Intended to be backwards compatible so order does not have to be set on articles.

Requires corresponding order property update in Contentful. 

CleanShot 2025-08-07 at 02 19 46@2x

CleanShot 2025-08-07 at 02 21 04@2x

### DIFF
--- a/src/content/content.types.ts
+++ b/src/content/content.types.ts
@@ -139,6 +139,7 @@ export type ArticleCategories = {
 type ArticleFields = {
   title: string
   id: string
+  order?: number
 }
 
 export type ArticleCategoryRaw = {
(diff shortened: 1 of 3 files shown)


tomer-tgp (approved):
You realest the number 9999 multiple types.
Can you maybe move it to a const ?
Not blocking

israelvicars:
@cursor In articleCategoriesTransformers, the value 9999 is set multiple times. Set it to a constant once and reference the constant.

PR #419, opened 2025-08-07
allow empty issues

tomer-tgp:
This PR fixes a bug - when a user deletes all their top issues, it was not saved to our DB.


israelvicars (approved):
🚢

PR #422, opened 2025-08-07
DT-188 - Fetch New Precomputed BR<>L2 Matches

oakinh:
https://goodpartyorg.postman.co/workspace/GP-API~66e2b59d-bf6a-4380-81c2-1cbf2d01bddf/run/38601784-7efca836-d091-499b-8ef8-f495a5a9532b

### DIFF
--- a/src/campaigns/campaigns.controller.ts
+++ b/src/campaigns/campaigns.controller.ts
@@ -279,6 +279,7 @@ export class CampaignsController {
         ...raceTargetDetails,
         electionType: L2DistrictType,
         electionLocation: L2DistrictName,
+        districtManuallySet: true,
       },
     })
   }
@@ -304,7 +305,7 @@ export class CampaignsController {
     const { district, winNumber, voterContactGoal, projectedTurnout } =
       raceTargetDetails
     const { L2DistrictType, L2DistrictName } = district
-    return await this.campaigns.updateJsonFields(campaign.id, {
+    return this.campaigns.updateJsonFields(campaign.id, {
       pathToVictory: {
         districtId: district.id,
         electionType: L2DistrictType,
@@ -315,6 +316,7 @@ export class CampaignsController {
         source: P2VSource.ElectionApi,
         p2vStatus: P2VStatus.complete,
         p2vCompleteDate: new Date().toISOString().slice(0, 10),
+        districtManuallySet: false,
       },
     })
   }
--- a/src/campaigns/campaigns.controller.ts
+++ b/src/campaigns/campaigns.controller.ts
@@ -1,4 +1,5 @@
 import {
+  BadRequestException,
   Body,
   ConflictException,
   Controller,
@@ -21,6 +22,7 @@ import { ElectionsService } from 'src/elections/services/elections.service'
 import { P2VStatus } from 'src/elections/types/pathToVictory.types'
 import { EnqueuePathToVictoryService } from 'src/pathToVictory/services/enqueuePathToVictory.service'
 import { PathToVictoryService } from 'src/pathToVictory/services/pathToVictory.service'
+import { P2VSource } from 'src/pathToVictory/types/pathToVictory.types'
 import { SlackService } from 'src/shared/services/slack.service'
 import { userHasRole } from 'src/users/util/users.util'
 import { ReqUser } from '../authentication/decorators/ReqUser.decorator'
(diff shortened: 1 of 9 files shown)


src/pathToVictory/services/pathToVictory.service.ts
```diff
@@ -149,22 +152,40 @@ export class PathToVictoryService extends createPrismaBase(
...
+          L2DistrictName: electionLocation,
+          electionDate: input.electionDate,
           state,
-          electionType,
-          electionLocation,
-          input.partisanType,
-          input.priorElectionDates,
-        )
+        })
+        if (!raceTargetDetails || !raceTargetDetails?.projectedTurnout) {
```
tomer-tgp, line 161: the first condition is not needed since you have a `?` after raceTargetDetails

src/pathToVictory/services/pathToVictory.service.ts
```diff
@@ -149,22 +152,40 @@ export class PathToVictoryService extends createPrismaBase(
...
+            electionLocation,
+            input.partisanType,
+            input.priorElectionDates,
+          )
 
-        if (counts?.total && counts.total > 0) {
-          pathToVictoryResponse.electionType = electionType
-          pathToVictoryResponse.electionLocation = electionLocation
-          pathToVictoryResponse.counts = counts
+          if (counts?.total && counts.total > 0) {
```
tomer-tgp, line 172: first condition is not needed

src/pathToVictory/services/pathToVictory.service.ts
```diff
@@ -149,22 +152,40 @@ export class PathToVictoryService extends createPrismaBase(
...
+        const { projectedTurnout, winNumber, voterContactGoal } =
+          raceTargetDetails
+
+        pathToVictoryResponse.electionType = electionType
+        pathToVictoryResponse.electionLocation = electionLocation
+        pathToVictoryResponse.counts = {
+          projectedTurnout,
+          winNumber,
+          voterContactGoal,
+        }
```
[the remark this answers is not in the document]
tomer-tgp: this is valid

src/pathToVictory/services/pathToVictory.service.ts
oakinh: I think in this case it is because we're doing a comparison operation
image

src/campaigns/campaigns.controller.ts
```diff
@@ -277,6 +279,44 @@ export class CampaignsController {
...
+    if (!raceTargetDetails) {
+      throw new InternalServerErrorException(
+        'Failed to fetch the raceTargetDetails for the requested campaign',
+      )
+    }
+    const { district, winNumber, voterContactGoal, projectedTurnout } =
+      raceTargetDetails
+    const { L2DistrictType, L2DistrictName } = district
+    return this.campaigns.updateJsonFields(campaign.id, {
+      pathToVictory: {
```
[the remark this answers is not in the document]
oakinh: updateJsonFields does a deepMerge on P2V though, so wouldn't me spreading P2V in here be redundant?
image

src/campaigns/campaigns.controller.ts
```diff
@@ -277,6 +279,44 @@ export class CampaignsController {
...
+        districtId: district.id,
+        electionType: L2DistrictType,
+        electionLocation: L2DistrictName,
+        winNumber,
+        voterContactGoal,
+        projectedTurnout,
+        source: P2VSource.ElectionApi,
+        p2vStatus: P2VStatus.complete,
+        p2vCompleteDate: new Date().toISOString().slice(0, 10),
```
RavenHursT, line 318: Looks like we have 2 _different_ ways that we format this date:

https://github.com/thegoodparty/gp-api/blob/14ca0eb28337579ea0ad8da4543f99c3bf4af7bf/src/admin/services/adminP2V.service.ts#L19
https://github.com/thegoodparty/gp-api/blob/bada25845757d30fa60aa9a9210e880d2afdde5f/src/pathToVictory/services/pathToVictory.service.ts#L410

Might be worth taking a look into whether or not we want to reconcile this change w/ these other two places where this value is set? 🤔

src/elections/services/elections.service.ts
```diff
@@ -82,11 +109,11 @@ export class ElectionsService {
     return projectedTurnout
       ? {
           ...this.calculateRaceTargetMetrics(projectedTurnout.projectedTurnout),
-          projectedTurnout: projectedTurnout.projectedTurnout,
           source: P2VSource.ElectionApi,
           electionType: projectedTurnout.L2DistrictType,
           electionLocation: projectedTurnout.L2DistrictName,
           p2vStatus: P2VStatus.complete,
+          p2vCompleteDate: new Date().toISOString().slice(0, 10),
```
RavenHursT, line 116: Same here.. we have a date formatter for ISO dates.. might be good to just use that.. 

https://github.com/thegoodparty/gp-api/blob/develop/src/shared/util/date.util.ts#L19-L24

src/pathToVictory/services/pathToVictory.service.ts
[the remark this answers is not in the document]
RavenHursT: This entire method is a cyclomatic complexity nightmare and really needs to be refactored.

I'm not suggesting to do this here and now, but I'm not surprised that control flow here is brittle and things are getting missed whenever someone has to go into this code to make changes.

RavenHursT (approved):
Non blocking comments.. just some things to consider..

tomer-tgp (approved):

PR #426, opened 2025-08-09
feat: public campaign endpoints

tomer-tgp:


israelvicars:
@tomer-tgp Review by GPT-5 is surfacing this blocking issue for name matching logic and a recommended fix. Seems non-redundant with the Bugbot feedback. Probably worth a look.

---

## Public Campaign Search (WEB-4414): Name matching and schema contract

### Summary
- New public endpoint `GET /public-campaigns` returns 404s for many valid queries due to overly strict slug matching.
- The response shape selected from Prisma does not fully align with the `FindByRaceIdResponse` Zod schema, creating a contract risk.

### Root causes
1) Name matching logic is too strict
   - Current implementation removes all dashes from the campaign slug and then splits on non-alphanumeric characters. Because slugs are typically alphanumeric with dashes, removing dashes collapses the slug into a single token, making equality checks almost never match partial names (e.g., last name only).
   - This regresses intended behavior for hyphenated names and common last-name searches.

2) Response schema mismatch
   - Zod response schema (`FindByRaceIdResponse`) expects a nested `website` object with multiple fields (e.g., `id`, `createdAt`, `campaignId`, `vanityPath`, etc.).
   - The Prisma `select` currently includes only `website.domain.name/status` and omits other fields referenced by the schema. If the DTO/schema is used for documentation or runtime validation, it will not match the selected shape.

### Required fix
1) Implement robust token-aware slug matching
   - Keep campaign slug tokens by splitting on `-`.
   - For last-name filtering: require an exact token match (case-insensitive, normalized) against any token in the campaign slug.
   - For first+last matching: require both normalized tokens to be present (either in order or at least contained as discrete tokens), and support hyphenated first or last names by splitting name inputs similarly.
   - Avoid substring `includes`-style matching that can produce false positives (e.g., `smith` matching `blacksmith`).

   Example approach (sketch):
   ```ts
   function normalizeToTokens(value: string): string[] {
     return slugify(value, { lower: true, strict: true })
       .split('-')
       .filter(Boolean);
   }

   function slugHasAllTokens(campaignSlug: string, requiredTokens: string[]): boolean {
     const campaignTokens = campaignSlug.split('-').filter(Boolean);
     const set = new Set(campaignTokens);
     return requiredTokens.every((t) => set.has(t));
   }

   // last-name only
   const lastTokens = normalizeToTokens(lastName);
   if (!slugHasAllTokens(campaign.slug, lastTokens)) return false;

   // then optionally require first-name tokens if provided
   const firstTokens = normalizeToTokens(firstName);
   if (firstTokens.length && !slugHasAllTokens(campaign.slug, firstTokens)) return false;
   ```

2) Align the response contract and selection
   - Option A (preferred): Expand the Prisma `select` to include the fields required by `FindByRaceIdResponse`'s `website` schema (e.g., `id`, `createdAt`, `updatedAt`, `campaignId`, `status`, `vanityPath`, `content`, and `domain { name, status }`). Ensure `campaignPositions` selection matches the schema as well.
   - Option B: Relax the Zod schema to the minimal subset actually returned. If choosing this option, rename the DTO to signal a partial/minimal response and update API docs accordingly.

3) Minor cleanups
   - Controller signature: since `NotFoundException` is thrown on null, return a non-nullable response type in the method signature for clarity.
   - Normalize `PublicAccess` import path for consistency with the rest of the codebase.

### Impacted files
- `src/campaigns/services/public-campaigns.service.ts`
- `src/campaigns/controllers/public-campaigns.controller.ts`
- `src/campaigns/schemas/public/FindByRaceIdResponse.schema.ts`

### Acceptance criteria
- Hyphenated and multi-token names (e.g., "Mary-Jane", "De La Cruz") are correctly matched.
- Last-name-only queries match only when the last name appears as a discrete token in the campaign slug; no false positives for substrings (e.g., `smith` does not match `blacksmith`).
- First+last queries prefer the exact candidate over other same-race candidates with the same last name.
- The response object returned by the endpoint matches the documented Zod schema, or the schema is updated to exactly match the returned shape.

israelvicars (approved):
:shipit:

PR #430, opened 2025-08-12
fix: handle createStep number/string type mismatch in website updates

Collin-GP:
Use z.coerce.string() to accept both number and string createStep values,
  fixing 400 validation errors that prevented website saves from working.

to test, boot up on develop, try to save websites, expect 400

check out this branch, expect to be able to save


Collin-GP:
bugbot run

RavenHursT (approved):

PR #435, opened 2025-08-13
Add better logging for electionApiGet

oakinh:


src/elections/services/elections.service.ts
```diff
@@ -85,6 +99,8 @@ export class ElectionsService {
       },
     )
 
+    console.log('positionWithDistrict: ', positionWithDistrict)
```
tomer-tgp, line 102: Same here, please use this.logger

tomer-tgp (approved):
please use this.logger and then we are good to merge

PR #447, opened 2025-08-15
Missing win numbers query fix

oakinh:


src/elections/services/elections.service.ts
```diff
@@ -235,7 +235,16 @@ export class ElectionsService {
...
+      .map((s) => s.trim())
+      .filter((s) => s.length > 0)
+    if (segments.length === 0) return L2DistrictName
+    let longest = segments[0]
+    for (let i = 1; i < segments.length; i++) {
+      if (segments[i].length > longest.length) {
+        longest = segments[i]
+      }
+    }
+    return longest
```
RavenHursT, line 248: Consider using regex and a reduce for this instead?

⸻

✅ Solution Using Regex + JavaScript

```typescript
const input = "This is one##And another##Final one##Supercalifragilisticexpialidocious##Tiny";
const regex = /(?:^|##)([^#]*)(?=##|$)/g;

const matches = [...input.matchAll(regex)].map(m => m[1]);
const longest = matches.reduce((a, b) => a.length >= b.length ? a : b);

console.log("Longest segment:", longest);
```

▶ Output:

Longest segment: Supercalifragilisticexpialidocious


⸻

🔍 Why This Works
	•	The regex extracts all segments between ##.
	•	map(m => m[1]) gets the captured group from each match.
	•	reduce(...) finds the longest by comparing string lengths.

PR #459, opened 2025-08-20
[WEB-4345] P2P endpoint for voter file upload

israelvicars:


israelvicars:
bugbot run

israelvicars:
bugbot run

src/peerly/services/p2pPhoneListUpload.service.ts
```diff
@@ -0,0 +1,143 @@
...
+
+      this.logger.log(
+        `P2P phone list uploaded successfully for campaign ${campaign.id}, token: ${token}`,
+      )
+
+      return { token, listName }
+    } finally {
+      // Clean up temporary voter file filter
+      await this.voterFileFilterService.delete(voterFileFilter.id)
+    }
```
RavenHursT, line 95: Whatever we end up doing here, we most def don't want to be writing a "temporary" record to the DB, that very likely won't be cleaned up in the case of error handlings.

wrapping whole chunks of code like this in `try-catch`es is also an anti-pattern and complicates error/exception handling. `try-catch`es should be targeted around succinct lines of code that we know will `throw` if an error/exception will be generated (i.e. `fetch` calls)

src/peerly/p2p.controller.ts
```diff
@@ -68,4 +74,35 @@ export class P2pController {
...
+    try {
+      const { token, listName } =
+        await this.p2pPhoneListUploadService.uploadPhoneList(campaign, request)
+
+      return {
+        success: true,
+        token,
+        listName,
+        message: 'Phone list uploaded successfully',
+      }
```
RavenHursT, line 93:
```suggestion
      return {
        token
      }
```

src/peerly/p2p.controller.ts
```diff
@@ -68,4 +74,35 @@ export class P2pController {
...
+
+      return {
+        success: false,
+        token: '',
+        listName: request.listName,
+        message:
+          error instanceof Error
+            ? error.message
+            : 'Failed to upload phone list',
+      }
```
RavenHursT, line 105:
```suggestion
      throw new BadGatewayException('Failed to upload phone list')
```

israelvicars:
bugbot run

PR #521, opened 2025-09-04
[WEB-4597] - Makes initials for request canvassers call come from the authenticated user instead of being hard-coded

RavenHursT:
https://goodparty.atlassian.net/browse/WEB-4597

image


israelvicars (approved):
:shipit:

PR #523, opened 2025-09-05
feat: CRUD for contacts segment

tomer-tgp:


src/contacts/contactsSegment/contactsSegment.controller.ts
```diff
@@ -0,0 +1,90 @@
...
+    @Param('id', ParseIntPipe) id: number,
+    @ReqCampaign() campaign: Campaign,
+  ) {
+    await this.contactsSegmentService
+      .findByIdAndCampaignId(id, campaign.id)
+      .then((segment) => {
+        if (!segment) {
+          throw new NotFoundException('Contacts segment not found')
+        }
+      })
```
RavenHursT, line 87: Do we _need_ to mix `async-await` w/ a promise pattern in these two cases?

Seems like `await` and then the conditional throw should suffice instead of creating new scope 🤔

RavenHursT (approved):

PR #524, opened 2025-09-06
feat: "simple" version of calling contacts

tomer-tgp:


src/contacts/services/contacts.service.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  async findContacts(
+    dto: ListContactsDTO,
+    campaign: CampaignWithPathToVictory,
+  ) {
+    if (!PEOPLE_API_URL) {
+      throw new BadGatewayException(
+        'PEOPLE_API_URL environment variable not configured',
+      )
+    }
```
israelvicars, line 32: Maybe error should throw on startup if this is required functionality? If so, maybe different placement of existence check and error. If not, seems ok here.

israelvicars (approved):
One possible change suggested.

PR #528, opened 2025-09-10
feat: add token

tomer-tgp:


src/contacts/services/contacts.service.ts
```diff
@@ -8,23 +8,31 @@ import { HttpService } from '@nestjs/axios'
...
     if (!PEOPLE_API_URL) {
       throw new BadGatewayException(
         'PEOPLE_API_URL environment variable not configured',
       )
     }
+    if (!PEOPLE_API_S2S_SECRET) {
+      throw new BadGatewayException(
+        'PEOPLE_API_S2S_SECRET environment variable not configured',
+      )
+    }
```
RavenHursT, line 35: The pattern we have elsewhere in the api is to have the check for critical env vars on file load, _outside_ of the class, and to just throw a generic `Error` instance.

That way the app is guaranteed to _not_ boot, and therefor won't pass health checks, and then the box won't be brought into rotation by ECS.

PR #541, opened 2025-09-17
Eng 3698 merge filters

tomer-tgp:


src/voters/voterFile/voterFile.controller.ts
```diff
@@ -127,4 +132,58 @@ export class VoterFileController {
...
+
+  @Put('filter/:id')
+  @UseCampaign()
+  async updateVoterFileFilter(
+    @Param('id', ParseIntPipe) id: number,
+    @Body() body: UpdateVoterFileFilterSchema,
+    @ReqCampaign() campaign: Campaign,
+  ) {
+    const filter: VoterFileFilter | null =
+      await this.voterFileFilterService.findByIdAndCampaignId(id, campaign.id)
```
RavenHursT, line 164: I don't get it.. 

If you have the ID of the filter from the path parameters of the controller, then why do we have to do a lookup by ID and camping ID? 

ID is a primary key. Is it not?

src/voters/voterFile/voterFile.controller.ts
```diff
@@ -127,4 +132,58 @@ export class VoterFileController {
...
+  @UseCampaign()
+  @HttpCode(HttpStatus.NO_CONTENT)
+  async deleteVoterFileFilter(
+    @Param('id', ParseIntPipe) id: number,
+    @ReqCampaign() campaign: Campaign,
+  ) {
+    const filter: VoterFileFilter | null =
+      await this.voterFileFilterService.findByIdAndCampaignId(id, campaign.id)
+    if (!filter) {
+      throw new NotFoundException('Voter file filter not found')
```
RavenHursT, line 185: We shouldn't need this check for delete operations.. deletes can just return no content regardless of whether the operation was successful or not

RavenHursT (approved):

PR #578, opened 2025-09-26
feat: add unknown filter

tomer-tgp:


stephtea (approved):
Looks fine, but @oakinh will likely need to merge this into his work as I think he's working on this function also.

PR #1260, opened 2026-03-09
feat: clerk integration

tomer-tgp:
WebApp PR
https://github.com/thegoodparty/gp-webapp/pull/1524

> [!NOTE]
> **High Risk**
> Replaces core request authentication and user identity linkage, and introduces a schema migration with a new unique constraint; regressions could lock users out or incorrectly associate accounts.
> 
> **Overview**
> **Migrates authentication to Clerk.** Replaces the global `JwtAuthGuard`/Passport strategies with a new global `ClerkSessionGuard` that verifies Clerk session tokens (and existing M2M tokens) and JIT-provisions/links a local `User` record via `clerkId`.
> 
> **Adds Clerk user sync + migration plumbing.** Introduces a `POST /webhooks/clerk` endpoint (Svix-verified via `CLERK_WEBHOOK_SECRET`) to create/link/update/soft-delete local users from Clerk events, adds `User.clerkId` with a unique index, and includes a bulk `scripts/clerk-user-migration.ts` to import existing users into Clerk and backfill `clerkId`.
> 
> **Removes legacy auth endpoints/features and adds a small profile sync API.** Drops `/authentication/register`, `/authentication/login`, social login, and admin impersonation, and adds `POST /users/sync-profile` to update `phone`/`zip`; tests/mocks are updated for the new `clerkId` field.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 081cadb392b3af54443d59b855c0dd31773b9509. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/authentication/services/clerk-webhook.service.ts
```diff
@@ -0,0 +1,159 @@
...
+        'Clerk user.created event missing email address',
+      )
+      return
+    }
+
+    const existingByClerkId = await this.usersService.findUser({
+      clerkId: data.id,
+    })
+    if (existingByClerkId) {
+      this.logger.info(
```
RavenHursT: All of these `.info` logs should prolly be `.debug`

RavenHursT (changes requested):

src/vendors/clerk/webhooks/clerk-events-handler.controller.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    switch (event.type) {
+      case CLERK_EVENT_USER_UPDATED:
+        await this.clerkEventsHandlerService.handleUserUpdated(event.data)
+        this.eventEmitter.emit(AUTH_USER_UPDATED_EVENT, authEvent)
+        break
+      case CLERK_EVENT_USER_DELETED:
+        await this.clerkEventsHandlerService.handleUserDeleted(event.data)
+        this.eventEmitter.emit(AUTH_USER_DELETED_EVENT, authEvent)
```
RavenHursT: What's the point of emitting the event if you're calling the service directly from here anyway? 🤔 

we should emit the _whole_ event so that any services server-side that need to consume said event, can do so w/o being tightly coupled to this module/controller

RavenHursT: Seems like, if identifying information for `User` (i.e. first/lastName, email, avatar, password) are all moving to Clerk, then we shouldn't need to sync anything locally and therefore shouldn't _need_ a webhook event handler at all, right?

RavenHursT (approved):

PR #1497, opened 2026-04-21
Fix cleanup-preview: drain ECS tasks and empty buckets before pulumi destroy

swain:
## Why

The scheduled Cleanup Preview workflow has failed every run for the past 4+ cycles (see [recent runs](https://github.com/thegoodparty/gp-api/actions/workflows/cleanup-preview.yml)). Stale preview stacks are accumulating in AWS — cost leak and resource clutter.

## Root cause

Two failure modes, both at the **Destroy Stack** step, after `pulumi destroy` starts:

1. **Every stale stack hits this:**
   ```
   aws:ecs:Cluster ecsCluster deleting (600s) error: ...
   ClusterContainsTasksException: The Cluster cannot be deleted while Tasks are active.
   ```
   Pulumi deletes the ECS Service, but the service's running tasks don't drain within the provider's 10-minute `DeleteCluster` wait, so the cluster delete fails.

2. **Stacks with CSV objects additionally hit:**
   ```
   aws:s3:Bucket tevyn-poll-csvs-bucket deleting (0s) error: ...
   BucketNotEmpty: The bucket you tried to delete is not empty
   ```
   `tevyn-poll-csvs-${stage}` and `zip-to-area-code-mappings-${stage}` are declared with `forceDestroy: false` in `deploy/index.ts`. Intentional for prod, but blocks cleanup for preview.

Flipping `forceDestroy` in Pulumi wouldn't help the already-stuck stacks (their state was written with `forceDestroy: false`), so the workflow handles both cleanups directly before calling `pulumi destroy`.

## Fix

New `Drain ECS tasks and empty preview buckets` step runs between the Clerk cleanup and the Pulumi destroy. For each stale stack it:

- scales the ECS service `desiredCount` to 0
- force-stops any remaining running tasks and waits for them to reach `STOPPED`
- empties the per-PR `tevyn-poll-csvs-pr-N` and `zip-to-area-code-mappings-pr-N` buckets

Everything is guarded with `|| true` / conditional checks so partially-destroyed stacks (missing cluster, missing bucket) don't explode the step.

---

> [!NOTE]
> **Medium Risk**
> Changes the automated preview cleanup workflow to actively scale down/stop ECS services and delete S3 objects before `pulumi destroy`, which can impact running preview environments if PR/stack identification is wrong or AWS calls behave unexpectedly.
> 
> **Overview**
> Fixes the `cleanup-preview` GitHub Actions workflow so stale preview stacks can be reliably destroyed.
> 
> Before running `pulumi destroy`, the workflow now drains ECS by scaling services to 0 and stopping/waiting for running tasks, and it empties the preview S3 buckets `tevyn-poll-csvs-pr-<N>` and `zip-to-area-code-mappings-pr-<N>` to prevent `ClusterContainsTasksException` and `BucketNotEmpty` failures during teardown.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7fd1d5cc5c97d7ab5b8171d91415ecdd495d45a1. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/cleanup-preview.yml
+++ b/.github/workflows/cleanup-preview.yml
@@ -9,7 +9,7 @@ on:
   workflow_dispatch:
   schedule:
     # Run every 3 hours
-    - cron: "0 */3 * * *"
+    - cron: '0 */3 * * *'
 
 jobs:
   discover-stale-stacks:
@@ -34,7 +34,7 @@ jobs:
       - name: Install Pulumi CLI
         uses: pulumi/actions@v6
         with:
-          pulumi-version: "^3"
+          pulumi-version: '^3'
 
       - run: npm ci
 
@@ -75,7 +75,7 @@ jobs:
       - name: Install Pulumi CLI
         uses: pulumi/actions@v6
         with:
-          pulumi-version: "^3"
+          pulumi-version: '^3'
 
       - run: npm ci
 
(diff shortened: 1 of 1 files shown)


swain (commented):
Review from Plannotator

**.github/workflows/cleanup-preview.yml:** No, wrong approach. We should just have forceDestroy: environment === 'preview' on any S3 buckets.

And, I don't think the ECS changes are needed. that resource should drain automatically when the delete fires...right? confirm that. And if not, take a similar approach to the S3 buckets. Then, put this up for PR, and notify me via FYI.

swain:
Closing in favor of #1501, which addresses the same cleanup-preview problem by making preview stacks inherently destroyable in Pulumi (forceDestroy on the two S3 buckets, forceDelete on the ECS service, faster deregistration/stop timeouts, wider cluster delete retry window) instead of draining from the workflow. Vanilla `pulumi destroy` succeeds for new preview stacks after that lands.

PR #1526, opened 2026-04-27
ENG-7494: Send campaign plan Slack message only on free→Pro upgrade

dealvz:
## ENG-7494: Send campaign-plan Slack message ONLY when a candidate upgrades to Pro

[ClickUp ticket](https://app.clickup.com/t/86ah1q6pe)

### Problem
The `#cas-clickup-tasks` Slack message was firing at plan generation time for all candidates regardless of tier. Free candidates who later upgraded to Pro were never re-surfaced, so their voter outreach never got executed.

### Solution
- **Plan generation:** Slack only fires when the campaign is already Pro at the time tasks are created.
- **Pro upgrade:** When a free candidate upgrades, Slack fires if they already have a plan.
- **Dedupe:** A `proUpgradeSlackNotifiedAt` timestamp is persisted in `campaign.details` so a downgrade + re-upgrade doesn't double-post. The write goes through `patchCampaignDetails` (Serializable transaction) to avoid a concurrent-write race.
- **Retry:** Slack sends retry up to 3 times with exponential backoff (500ms / 1s / 2s), then fail silently so a Slack outage can't break the upgrade flow.

### Files changed
- `prisma/schema/campaign.jsonTypes.d.ts` + `updateCampaign.schema.ts` — added `proUpgradeSlackNotifiedAt` to the `CampaignDetails` type and Zod schema.
- `campaignTasks.service.ts` — gated plan-gen notification on `isPro`; added `notifySlackOnProUpgrade` with dedupe + retry; extracted shared `sendCampaignPlanSlackMessage` helper.
- `campaigns.service.ts` — injected `CampaignTasksService` and calls `notifySlackOnProUpgrade` on the free→Pro transition.
- Tests updated to cover all new paths (113/113 passing).

### No migration needed
`proUpgradeSlackNotifiedAt` is stored inside the existing `details` JSON column — no DDL changes.

### DIFF
--- a/prisma/schema/campaign.jsonTypes.d.ts
+++ b/prisma/schema/campaign.jsonTypes.d.ts
@@ -29,6 +29,7 @@ declare global {
       runForOffice?: 'yes' | 'no' | null
       pledged?: boolean
       isProUpdatedAt?: number // TODO: make this an ISO dateTime string
+      proUpgradeSlackNotifiedAt?: number
       customIssues?: Record<'title' | 'position', string>[]
       runningAgainst?: Record<'name' | 'party' | 'description', string>[]
       geoLocation?: {
--- a/src/campaigns/schemas/updateCampaign.schema.ts
+++ b/src/campaigns/schemas/updateCampaign.schema.ts
@@ -26,6 +26,7 @@ const CampaignDetailsSchema = z
     runForOffice: z.enum(['yes', 'no']),
     pledged: z.boolean(),
     isProUpdatedAt: z.number(),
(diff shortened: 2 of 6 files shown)


tomer-tgp:
Task linked: [ENG-7494 Send campaign plan slack message ONLY when a candidate upgrades to Pro](https://app.clickup.com/t/90132012119/ENG-7494)

feliksalbright (approved):
Sorry, didn't realize this one was sitting!

LGTM.

PR #1530, opened 2026-04-28
Make test user cleanup clerk-aware

swain:
The previous per-PR Clerk cleanup ran as an ECS task during preview teardown — fragile, easy to lose track of, and doing nothing for `dev`/`qa` users that piled up between runs.

Folding the work into the existing `UsersService.deleteTestUsers` scheduled job means cleanup just *happens* on a 6h interval everywhere gp-api runs, with the same throttling and error semantics as the rest of our Clerk integration. Postgres is per-env so the DB half stays scoped naturally; the Clerk half walks the shared dev pool, but since it only matches `@test.goodparty.org` and only touches users older than 3h, the blast radius is bounded to the test user domain we own — well past the lifetime of any in-flight e2e run.

---

> [!NOTE]
> **Medium Risk**
> Adds automated deletion of Clerk accounts for `@test.goodparty.org` users, which is an external-system destructive operation and could remove unintended accounts if filtering/cutoff logic is wrong. Otherwise changes are scoped to test-domain users and include throttling and per-user error handling.
> 
> **Overview**
> Moves test-user cleanup from preview teardown into the app’s scheduled `UsersService.deleteTestUsers` job by making it **Clerk-aware**: it now deletes stale `@test.goodparty.org` users from both Postgres and Clerk (older than 3 hours), using `clerkThrottle` and per-user try/catch logging.
> 
> Removes the per-preview ECS-based “Delete Clerk users” step from the `cleanup-preview` GitHub Actions workflow, leaving the workflow focused on Pulumi stack discovery/destruction (plus minor YAML quoting tweaks).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 97360c35a0d3013219622d6aacebd261ec711e13. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/cleanup-preview.yml
```diff
@@ -75,56 +75,10 @@ jobs:
       - name: Install Pulumi CLI
         uses: pulumi/actions@v6
         with:
-          pulumi-version: "^3"
+          pulumi-version: '^3'
 
       - run: npm ci
 
-      - name: Delete Clerk users
```
swain, line 82: This task fails (i.e. doesn't work) and causes previews to not be cleaned up automatically. Removing it solves multiple problems

stephtea (approved):

oakinh (approved):

PR #1543, opened 2026-05-04
ENG-7570 fix Pro renewal checkout

dealvz:
## Summary
- Block Pro checkout creation when the campaign election date is missing or in the past so users are not sent to Stripe for a subscription that cannot activate cleanly.
- Let paid Stripe subscription checkout webhooks complete Pro activation even when the election date is stale, preventing users from remaining stuck in Subscription Pending after money changes hands.
- Add an admin-only recovery endpoint for already-stuck users and focused tests for checkout validation, webhook behavior, and replay safety.

## Root Cause
The subscription checkout webhook previously threw after payment when the campaign election date was missing or in the past. That prevented `subscriptionId`, `isPro`, and `checkoutSessionId` cleanup from being persisted.

## Validation
- `npm test -- src/payments/purchase.controller.test.ts src/payments/services/paymentEventsService.test.ts`
- `npx eslint src/payments/payments.controller.ts` (existing max-params warning only)

### DIFF
--- a/src/payments/payments.controller.ts
+++ b/src/payments/payments.controller.ts
@@ -10,9 +10,11 @@ import {
   Post,
   RawBodyRequest,
   Req,
+  UseGuards,
 } from '@nestjs/common'
 import { UserRole } from '@prisma/client'
 import { Roles } from 'src/authentication/decorators/Roles.decorator'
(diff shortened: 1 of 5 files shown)


tomer-tgp:
Task linked: [ENG-7570 Cannot Renew Pro for 2026 Campaign](https://app.clickup.com/t/90132012119/ENG-7570)

PR #1641, opened 2026-05-25
Prevent Clerk identity rebinding on existing users

selta-goodparty:
## Summary

Adds validation to prevent rebinding a Clerk identity to a user account that already has an associated Clerk ID. This protects against accidental or malicious identity conflicts during the user linking flow.

## Changes

- **Added rebind protection**: When linking a user by email, check if the existing user already has a `clerkId`. If so, log a warning and return `null` instead of proceeding with the update.
- **Improved logging**: Clarified the log message from "Linking existing user" to "Linking legacy user" to better reflect the scenario of connecting legacy accounts to Clerk for the first time.
- **Added context to warning**: The warning log includes the existing Clerk ID and incoming Clerk ID for debugging identity conflicts.

## Implementation Details

The check is placed early in the existing email-lookup flow, before any database updates occur. This ensures we fail safely and log the conflict for investigation without modifying user state.

https://claude.ai/code/session_014bW8faFSiYejrbdUCtDv5o

---

> [!NOTE]
> **Medium Risk**
> Touches Clerk provisioning in the session path, but the change is a fail-closed guard that prevents identity takeover rather than broadening access.
> 
> **Overview**
> **`findOrProvisionByClerk`** no longer overwrites **`clerkId`** when an email match finds a user who already has a Clerk identity. It logs a warning (user id, existing vs incoming Clerk id), returns **`null`**, and skips the DB update.
> 
> First-time email linking for users **without** a **`clerkId`** is unchanged; the info log text is clarified to **"Linking legacy user to Clerk account"**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7a5393f76dba8c6d21fb708b05ec332b250d9c8d. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/users/services/users.service.test.ts
+++ b/src/users/services/users.service.test.ts
@@ -424,133 +424,80 @@ describe('UsersService', () => {
   })
 
   describe('findOrProvisionByClerk', () => {
+    const provision = (
+      clerkId: string,
+      email: string,
+      first = 'Test',
+      last = 'User',
+    ) =>
+      usersService.findOrProvisionByClerk({
+        clerkId,
+        email,
+        firstName: first,
+        lastName: last,
+      })
+
+    const createUser = (email: string, clerkId: string | null = null) =>
+      service.prisma.user.create({
+        data: { email, clerkId },
+      })
+
     it('returns existing user when found by clerkId', async () => {
-      const existing = await service.prisma.user.create({
-        data: {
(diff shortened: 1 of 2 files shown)


selta-goodparty:
delegate review

PR #1645, opened 2026-05-26
Prevent prototype pollution in multipart form handling

selta-goodparty:
## Summary

Add protection against prototype pollution attacks in the `FilesInterceptor` by blocking assignment to forbidden keys (`__proto__`, `constructor`, `prototype`) in both direct form fields and nested object paths.

## Changes

- **Forbidden keys set**: Created `FORBIDDEN_KEYS` constant containing `__proto__`, `constructor`, and `prototype`
- **Direct field filtering**: Updated the multipart form value assignment to skip forbidden keys via `else if (!FORBIDDEN_KEYS.has(part.fieldname))`
- **Nested path validation**: Added two guards in `setNestedProperty()`:
  - Check for forbidden keys in simple (non-nested) paths
  - Check for forbidden keys anywhere in the split path array for nested assignments
- **Code formatting**: Reformatted multi-line conditionals for readability (no logic change)

## Implementation Details

The protection is applied at two levels:
1. When assigning form fields directly to `req.body`
2. When setting nested properties via bracket notation (e.g., `obj[key][subkey]`)

This prevents attackers from polluting the prototype chain through either direct field names or nested path components.

https://claude.ai/code/session_01Sc4ow1GZrVnwr9wSQr4TDF

---

> [!NOTE]
> **Medium Risk**
> Security fix on a shared request interceptor that shapes `req.body` for all multipart uploads; behavior change only for malicious or forbidden key names, with no new tests in the diff.
> 
> **Overview**
> Hardens **`FilesInterceptor`** so multipart field names cannot pollute `Object.prototype` when values are merged into `req.body`.
> 
> A local **`FORBIDDEN_KEYS`** set (`__proto__`, `constructor`, `prototype`) now blocks **top-level** field assignment unless the name is allowed, and **`setNestedProperty`** silently skips writes when any path segment (including simple keys) matches that set. Unrelated formatting-only tweaks to multi-line `if` conditions in the same helper.
> 
> This closes the gap where bracket notation (e.g. `foo[__proto__][polluted]`) could previously set dangerous keys on nested objects built from form data.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7e4094cb657833ec86423b380e4d944e8a64dab1. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/campaigns/ai/schemas/forbiddenKeys.ts
+++ b/src/campaigns/ai/schemas/forbiddenKeys.ts
@@ -1,1 +1,1 @@
-export const FORBIDDEN_KEYS = new Set(['__proto__', 'constructor', 'prototype'])
+export { FORBIDDEN_KEYS } from '@/shared/constants/forbiddenKeys'
--- a/src/files/interceptors/files.interceptor.test.ts
+++ b/src/files/interceptors/files.interceptor.test.ts
@@ -1,0 +1,128 @@
+import { CallHandler, ExecutionContext } from '@nestjs/common'
+import { describe, expect, it, vi } from 'vitest'
+import { of } from 'rxjs'
+import { Headers, MimeTypes } from 'http-constants-ts'
+import { FilesInterceptor, setNestedProperty } from './files.interceptor'
+
+const FORBIDDEN = ['__proto__', 'constructor', 'prototype'] as const
+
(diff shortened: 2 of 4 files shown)


selta-goodparty:
delegate review

